PR #141, opened 2011-09-15
fix default date format for catalan

oriolgual:


thibaudgg:
Are the dashes separation correct? According to https://en.wikipedia.org/wiki/Date_format_by_country, it should use slashes like for Spain.

PR #502, opened 2014-10-30
Fix confirmation error message for Italian.

mperrando:
The errors.messages.confirmation string is still obsolete, without the
%{attribute} parameter.


mperrando:
Rebased just now on the master branch.

Hope it can be merged: it's been waiting since October 2014.

PR #504, opened 2014-11-20
Fix "i18n" gem dependency

exAspArk:
Hi, 

I have an error when I use `i18n 0.7.0.beta1` gem in my Gemfile with `rails 3.2.21`:
```
cannot load such file -- i18n/core_ext/string/interpolate
.../ruby/2.1.0/gems/activesupport-3.2.21/lib/active_support/core_ext/string/interpolation.rb:2:in `require'
```

To fix it I suggest to lock version dependency below `0.7.0.beta1`.


digitalfrost:
Rails 3.2.21 is no longer supported.
Perhaps this ticket should be closed?

PR #509, opened 2014-12-15
Update de.yml: human.format.precision should be 3

salimhb:
according to rails doc: http://apidock.com/rails/v4.1.8/ActionView/Helpers/NumberHelper/number_to_human_size


jhilden:
thanks for merging this

PR #512, opened 2014-12-31
Update el.yml

panos78:
I have updated the greek translation.
:-)


kuroda:
@panos78 I can't merge because the Travis CI build failed. But, I don't see anything wrong in your commit.

I will work on it when I have some leisure time. Sorry for delay.

PR #517, opened 2015-02-07
Changed Turkish currency unit symbol.

computeus:


kuroda:
@computeus 

On my browser, this character is rendered like `☒`. So, I am hesitant to accept your pull request.

computeus:
@kuroda 

According to http://www.unicode.org/versions/Unicode6.2.0/ the Turkish Lira Sign is addded to Unicode.

kuroda:
@computeus 

According to http://stackoverflow.com/questions/14117547/browser-support-and-workarounds-for-turkish-lira-sign (a post in 2013), few fonts support Turkish Lira Sign and one answer says

> it has usually taken about ten years or more from the adoption of a character into Unicode to its general availability in fonts.

My opinion is that we should wait for a while.

digitalfrost:
Can be closed?

PR #521, opened 2015-02-19
added tatar language locale file

fattakhov:
Created locale yml file for tatar language: http://en.wikipedia.org/wiki/Tatar_language


kuroda:
Great!

PR #523, opened 2015-02-23
Remove exclamation marks

tjku:
This changes are done by [Psych](https://rubygems.org/gems/psych) (version 2.0.13).

List of changes:
- add `---` mark (to signal the start of a document if no directives are present)
- comments are removed
- replace single quotation marks with double quotation marks
- add space to empty keys. I don't know if this is necessary, but I'll leave it as a part of automated Psych parsing


kuroda:
@tjku Thanks!

PR #524, opened 2015-02-25
Fix Croatian currency format

toncid:
It was invalid: it used decimal point instead of comma.


kuroda:
@toncid Could you squash your commits into one?

toncid:
Done: #525

PR #525, opened 2015-02-25
[hr.yml] Update currency format

toncid:
The Croatian locale did not provide the correct currency format (decimal comma and period as the thousands' delimiter).


toncid:
Thanks! When can we expect a gem update?

PR #526, opened 2015-02-26
Sorting locales

tjku:
In my opinion, locales should be sorted by keys. This allows to unify different locales, so they are easier to maintain them. Also this gives rule, how and where add new keys in the future.

For example, other projects which do this: [devise-i18n](https://github.com/tigrish/devise-i18n) and [will-paginate-i18n](https://github.com/tigrish/will-paginate-i18n).

To sort keys I used gem [i18n_yaml_sorter](https://github.com/redealumni/i18n_yaml_sorter).


kuroda:
@tjku Basically, it is OK. But, before accepting your commits, I want to hear from other committers.

You think we should place `billion` before `million`, don't you?

I am hesitant because the Rails team adopts *semantic* ordering system. See below:

- https://github.com/rails/rails/blob/master/activesupport/lib/active_support/locale/en.yml

tjku:
@kuroda Yes, `billion` before `million`.

But I see your point. If alphabetic order is not agreeable with other committers, I suggest other sorting rule (e.g. semantic).

My point is to be consistent around all locales. Now there are some locales sorted alphabetically, others are sorted semantically, and others aren't sorted at all.

kuroda:
@tjku After discussions with one of friends, my conclusion is that many committers will not love the *alphabetical* ordering system like that of devise-i18n.

I think we should write a script to sort keys *semantically* and *consistently*.

I will try when I find some leisure time.

Of course, I will also welcome *your* script :wink:

digitalfrost:
For me it makes sense for rails-i18n to mirror rails, so we should stick with semantic ordering.

PR #542, opened 2015-04-30
Spanish does not have Serial comma

jguitar:
Hi,

Spanish does not have [Serial comma](http://en.wikipedia.org/wiki/Serial_comma). You can take a look to [RAE](http://lema.rae.es/dpd/?key=coma2&lema=coma2) (note: is in Spanish)


jguitar:
And it was strange because there are some spanish (like `'es-MX'`) without the comma

PR #551, opened 2015-05-19
Update de.yml

reditec:


fphilipe:
@kuroda Why was this merged? This is wrong. This basically changed plain durations into *how long ago* expressions. Could you revert this please?

PR #558, opened 2015-06-23
fix SI byte values, especially 'kB' with lowercase 'k'

themilkman:
This PR replaces non-standard conform Byte-values with those standardized by SI, see: https://en.wikipedia.org/wiki/Byte#Unit_symbol
This will allow standard conform parsing of those values.

### DIFF
--- a/rails/locale/af.yml
+++ b/rails/locale/af.yml
@@ -10,7 +10,7 @@ af:
     - Vry
     - Sat
     abbr_month_names:
-    - 
+    -
     - Jan
     - Feb
     - Mar
@@ -36,7 +36,7 @@ af:
       long: "%B %d, %Y"
       short: "%b %d"
     month_names:
-    - 
+    -
     - Januarie
     - Februarie
     - Maart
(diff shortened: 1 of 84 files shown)


kuroda:
@themilkman 

Thanks for your pull request. I can't accept it immediately, though.

The purpose of our repository is just to gather the translation data of Rails _official_ locale files.
Currently, the Rails adopts `KB` for the `number.human.storage_units.units.kb` key. 
So, you should send a pull request to the `rails/rails` repository itself before we adopt the `kB`.

See https://github.com/rails/rails/blob/master/activesupport/lib/active_support/locale/en.yml

Please let us know when this change happens on the rails repository.

themilkman:
@kuroda
Thanks for your quick response!  I'll create a PR for the original locale file during the next days and will inform you about any news.

themilkman:
Since the rails PR was rejected (without any discussion) I'll close this PR an leave it (imho as incorrect) as it is.
Anyway, thanks for your response and for managing this helpful project in general!

kuroda:
@themilkman I'm sorry for the rejection and thank you for your intention to improve this unperfected world.

PR #565, opened 2015-07-12
remove full stops from ja

marutosi:
http://www.redmine.org/issues/20296

It is strage only Japanese has full stops.
Simplified and Traditional Chinese do not have these.

### DIFF
--- a/rails/locale/ja.yml
+++ b/rails/locale/ja.yml
@@ -99,38 +99,38 @@ ja:
   errors:
     format: "%{attribute}%{message}"
     messages:
-      accepted: を受諾してください。
-      blank: を入力してください。
-      present: は入力しないでください。
-      confirmation: と%{attribute}の入力が一致しません。
-      empty: を入力してください。
-      equal_to: は%{count}にしてください。
-      even: は偶数にしてください。
-      exclusion: は予約されています。
-      greater_than: は%{count}より大きい値にしてください。
-      greater_than_or_equal_to: は%{count}以上の値にしてください。
-      inclusion: は一覧にありません。
-      invalid: は不正な値です。
-      less_than: は%{count}より小さい値にしてください。
-      less_than_or_equal_to: は%{count}以下の値にしてください。
-      not_a_number: は数値で入力してください。
-      not_an_integer: は整数で入力してください。
-      odd: は奇数にしてください。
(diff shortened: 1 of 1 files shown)


amatsuda:
Not that I'm strongly opposing to this change, but to me, "Simplified and Traditional Chinese do not have these" didn't sound like a compelling reason for the change.
So let me just describe the historical background.

[The initial version of the Japanese translations](900c6b67a1c2eaabdbed2e883cd84861556c49e9) are taken from the Ruby-GetText-Package project, and in my understanding, the basic policy is:
* sentences should end with a `。`
* words should not end with a `。`

which seems quite natural as a general Japanese writing rule.

OTOH the Chinese locale didn't have any full stops [from the beginning](21975ce1e227a6bf4efeae7556d5c24b72e9ed7a). I don't know much about the Chinese language, but maybe they're not so uncomfortable with sentences without full stops?

Anyway, it's not at all "strange" that ja and zh-* differs. These are for different cultures . These are in fact different from the beginning of this project, and I think we both have been happy with each locale's own rule so far.
Aren't there any other reason than "Chinese locales are not like this"? I'd like to see something like "as a Japanese, it's seems more natural not to include full stops because ...".

marutosi:
@amatsuda, your links are broken. Please fix these.

marutosi:
Not only "Simplified and Traditional Chinese" but **all** languages do not have full stop at **empty**.
It is strange **only** Japanese has full stop at **empty**.

amatsuda:
@marutosi Oops! Fixed! (I guess we hit a GH :bug:...)

marutosi:
Original claim is http://www.redmine.org/issues/20296 .
> 「メンバーの保存に失敗しました: ロールを入力して下さい。.」

Double full stop "。." is strange.

http://www.redmine.org/attachments/13932/Fix-typo-japanese-translate.patch
> notice_failed_to_save_members: "メンバーの保存に失敗しました: %{errors}."

It is strange **only** Japanese has full stop at **error**.

amatsuda:
All right. I understand that you value at consistency with other locales.

I just described another underlying logic, consistency with other I18n packages in the history, which probably you might not taken care of.

Some people might not like the change because of their point of view, but if you're still strongly confident with your logic, I'm not gonna fight.
As I stated at the very beginning, I have no preference for this particular matter. I just wanted to let people here know what I know.

PR #569, opened 2015-07-28
correct Japanese translation.

tbrisker:


kuroda:
As a native speaker, I think these translations are too colloquial.

PR #571, opened 2015-08-31
add support for greek transliteration

atzorvas:


kuroda:
@atzorvas Thanks.

Could you squash two commits into one? You can do it with like this:

    git rebase -i HEAD~2
    git push -f

atzorvas:
@kuroda done, I think

PR #576, opened 2015-09-05
Update el.yml correcting @panos78's submitted file

digitalfrost:
Fixes #512


teoulas:
The `delimiter` and `separator` characters are wrong:

https://en.wikipedia.org/wiki/Decimal_mark#Hindu.E2.80.93Arabic_numeral_system

PR #586, opened 2015-10-15
Update de-CH to latest de-locales

Elektron1c97:


rails/locale/de-CH.yml
```diff
@@ -56,45 +56,45 @@ de-CH:
...
+        one: vor einer Minute
+        other: "vor %{count} Minuten"
       x_months:
-        one: ein Monat
-        other: "%{count} Monate"
+        one: vor einem Monat
+        other: "vor %{count} Monaten"
       x_seconds:
-        one: eine Sekunde
-        other: "%{count} Sekunden"
```
fphilipe: This new part is wrong. The old one is the correct way. See #595 and #551.

Elektron1c97: ok, I will change it

Elektron1c97: @fphilipe Updated

PR #587, opened 2015-10-15
Use turkish lira symbol instead of TL for TR locale

fotos:
Since 2012 there is a symbol (`₺`) for the [Turkish Lira](https://en.wikipedia.org/wiki/Turkish_lira_sign), introduced in Unicode 6.2. Most locales have a symbol in `number.currency.format.unit` and I think the turkish locale should follow suit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,26 @@ Following locales have some missing translations or pluralizations:
 
 We always welcome your contributions!
 
+## Currency symbols
+
+Some locales have the symbol of the currency (e.g. `€`) under the key `number.currency.format.unit`,
+while others have the code (e.g. `CHF`). The value of the key depends on the widespread adoption of
+the unicode currency symbols by fonts.
+
+For example the Turkish Lira sign (`₺`) was recently added in Unicode 6.2 and while most popular
+fonts have a glyph, there are still many fonts that will not render the character correctly.
+
+If you want to provide a different value, in a Rails app, you can create your own locale file under
+`config/locales/tr.yml` and override the respective key:
+
+```YAML
+tr:
+  number:
+    currency:
+      format:
+        unit: TL
+```
+
 ## How to contribute
 
 ### Quick contribution
--- a/rails/locale/tr.yml
+++ b/rails/locale/tr.yml
@@ -142,7 +142,7 @@ tr:
         separator: ! ','
         significant: false
         strip_insignificant_zeros: false
-        unit: TL
+        unit: ₺
     format:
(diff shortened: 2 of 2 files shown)


kuroda:
We have discussed on this in #517. I need an evidence that _most_ locales including non-European locales have this symbol.

digitalfrost:
I've tweeted to @johnholtripley to ask if he could test support for U+20BA ₺ turkish lira sign across browsers and devices on Unify (a tool for determining the support of Unicode characters  http://unicode.johnholtripley.co.uk/ )

digitalfrost:
A lot of main stream fonts now seem to support U+20BA ₺ turkish lira sign

http://www.fileformat.info/info/unicode/char/20ba/fontsupport.htm

digitalfrost:
Testing font support in https://www.google.com/fonts by inputing ₺ in the preview text box seems to indicate widespread font support for this unicode character

fotos:
@kuroda thanks for pointing that discussion out. I did a search on PRs / issues but apparently I missed that. Anyway, I wanted to see the current support of Google Web fonts. @digitalfrost beat me to that. :smile:

Now that I found a bit of time I wrote a [small script](https://gist.github.com/fotos/fea51366458ed4b1a189) that checks how many of the top 100 most popular fonts (out of 732) have the glyph using [TTFunk](https://github.com/prawnpdf/ttfunk).

Here are the results:

|                | Glyph | No Glyph | Unknown |
| ----------- | -------- | ------------ | ------------- |
| Top 100 | 53      | 46            | 1              |
| Total      | 278    | 444          | 5              |

Now all that's left is to answer if they are a sufficient percentage. I'll leave that to you. :smile:

fotos:
Another idea would be to have 2 keys, one for the unit and another for the symbol. This way we could have both the symbol and the unit for all locales wherever they are available (e.g. for euro `unit: EUR` and `symbol: €`). Unfortunately this might break existing apps. Just a thought…

kuroda:
@fotos Thanks a lot. I am getting inclined to accept your PR.

Could you add a note on the Turkish lira sign to the README.md as a separate section below "Available Locales"? In fact, there are many fonts that don't support this glyph still. We should provide an instruction to override the default configuration.

fotos:
@kuroda done.

Please review for any spelling / grammar mistakes. I hope the new section clarifies the situation and is helpful and meaningful for overriding the defaults.

kuroda:
Great!

PR #589, opened 2015-10-21
correct dutch translations

dennisvandehoef:
For example, the translation from "trillion" was, that what in english is a "billion".

### DIFF
--- a/rails/locale/nl.yml
+++ b/rails/locale/nl.yml
@@ -10,7 +10,7 @@ nl:
     - vr
     - za
     abbr_month_names:
-    - 
+    -
     - jan
     - feb
     - mrt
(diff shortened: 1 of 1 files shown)


florish:
@dennisvandehoef Thanks for the contribution! I'm not sure whether your corrections are actually, well, correct :) :

* The English [quadrillion](https://en.wikipedia.org/wiki/Quadrillion) is 10^15, which in Dutch is actually called [biljard](https://nl.wikipedia.org/wiki/Biljard).
* Same goes for [trillion](https://en.wikipedia.org/wiki/Trillion), which is 10^12. In Dutch this is [biljoen](https://nl.wikipedia.org/wiki/Biljoen)

If you agree, can you please revert these changes? Thanks!

florish:
@dennisvandehoef I opened a separate PR for the fix, please ignore my previous revert request.

PR #590, opened 2015-10-23
Fix awkward Khmer translations

keang:


keang:
@kuroda Please help to merge these translation fixes =D

digitalfrost:
@keang Lines 81-89  - why did you remove the **one** and **other** keys ?

keang:
@digitalfrost when `count=1` they are identical. There's no plural form..

digitalfrost:
@keang - that was what I gathered - thank you for confirming.

I was concerned that removing some of the keys would affect 
```
bundle exec rake i18n-spec:completeness rails/locale/en.yml rails/locale/km.yml
```
but looking at the code of **i18n-spec** - it handles this type of differences in pluralization keys...
I will merge your changes

PR #596, opened 2015-11-12
Fixed thousands separator and decimal mark for Greek.

teoulas:
Fixed thousands separator and decimal mark for Greek.

In Greece this is correct: "10.000,50 €"
This isn't: "10,000.50 €"
https://en.wikipedia.org/wiki/Decimal_mark#Hindu.E2.80.93Arabic_numeral_system

### DIFF
--- a/rails/locale/el.yml
+++ b/rails/locale/el.yml
@@ -141,17 +141,17 @@ el:
   number:
     currency:
       format:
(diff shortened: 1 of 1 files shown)


teoulas:
This PR reverts part of #576 and #512. The correct characters were used before.

PR #601, opened 2015-11-24
Updated pa locale to accommodate changes recommended by @Gurpartap

punjab:
Update to pa locale to accommodate changes recommended by @Gurpartap on diff a7d33d1


kuroda:
Merged. Thanks.

PR #611, opened 2015-12-26
comment rails 4.0 in readme

kbrock:
[skip ci]

Looks like some commits that went in now have master pegged to rails 5.x

Updated the readme to reflect


kbrock:
thanks for the great gem

PR #616, opened 2016-01-08
Changed Norwegian time separator from colon to dot.

espen:
Both are allowed since new changes in 2014 but dot is preferred and has been the standard.

Source (Norwegian): http://www.korrekturavdelingen.no/K4Klokkeslett.htm

### DIFF
--- a/rails/locale/nb.yml
+++ b/rails/locale/nb.yml
@@ -203,7 +203,7 @@ nb:
   time:
     am: ''
     formats:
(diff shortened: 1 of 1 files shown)


espen:
I'm having second thoughts about this. I think there are advantages of using colon and I see that some big national sites are using that for data display and forms. Though in natural text it's probably more suitable to use dot.

I'm gonna close this for now but feel free to discuss. Both styles are valid as mentioned.

Issue #609, opened 2015-12-23
missing interpolation argument :attribute


TravisTam:
I have set my locale to zh-HK and seems that is difference from using zh-TW. It shows below errors. Can many one advise how can I fix it?
```
missing interpolation argument :attribute in "%{attribute}驗證失敗︰%{errors}" ({:errors=>"Nickname是無效的"} given)
```

TravisTam:
in the locale file, it is found %{attribute} is no longer existed in both 'en.yml' and 'zh-TW.yml'. I would like to know Is there any special setup I need to do to satisfy the zh-HK locale setting?
```
zh-HK:
   errors:
      messages:
         record_invalid: "%{attribute}驗證失敗︰%{errors}"
```

```
zh-TW:
   errors:
      messages:
         record_invalid:'校驗失敗: %{errors}'
```
```
en:
   errors:
      messages:
         record_invalid:'  'Validation failed: %{errors}'
```

kuroda:
@TravisTam I think we should remove `%{attribute}` from the translation.

I found many such arguments within `zh-HK` and `zh-YUE` locales. It would be very helpful if you make changes and send us a pull request.

PikachuEXE:
I got the same thing too :S

PR #621, opened 2016-01-27
! Update translations entries for zh-HK & zh-YUE with format similar to zh-TW

PikachuEXE:
This is trying to avoid some error like "missing interpolation argument :attribute"

Hopefully will solve #609


PikachuEXE:
Hello anyone?

PR #622, opened 2016-01-28
writing translation that would fit all arabic tens,gender rules, and …

mshka:
hello,
i should start by explaining a little about arabic and its many grammar "rules",
arabic unlike english has 6 rules of plurality which is already defined and used in this gem but with it come rules of gender which is not, for example the verb "be(كان)" in the present tense for masculine its (يكون) but for feminine its (تكون) and for nouns like "blank(فارغ)" in masculine its (فارغ) but for feminine its (فارغة) and for some plurals its (فارغة) in ur gem the already defined expression for e.g: blank:(فارغ، يرجى ملء الفراغ)  is correct for email(البريد الالكتروني) which is masculine but not for password(كلمة السر) which is a feminine noun and many other, plus in arabic we don't always use the attribute in the beginning of the message sometime it comes in the middle sometime in the very end

 so in my pull request along with a translator help we defined all error messages in a way where all tens and gender rules would be applied without any problem along with the correct format plus the old month names are barely used nowadays so i replaced them with the new ones and added the abbreviations in addition to  changing the long date format which is reversed for arabic 

thank you in advance


kuroda:
@mshka Thanks for your contribution!

Currently, your `ar.yml` has syntax errors on the line 163 and 200.

Adding quotation marks to the translation may solve your problem.

Please run `bin/rspec` and make sure all tests passed before submitting another pull request.

mshka:
hello @kuroda, i fixed it :grin:
thank you

kuroda:
Done. 69440fca7f92945065ce845befb108555797a9ce

PR #627, opened 2016-02-04
Added support for malayalam language(iso code - ISO 639-1)

deepak07:
- Added support for malayalam language(iso code - ISO 639-1)
  [https://en.wikipedia.org/wiki/Malayalam](url)


deepak07:
@kuroda Can you look into this PR...

PR #651, opened 2016-05-27
Don't translate most of storage units in ja locale

sakuro:
In my personal experience, Japanese people are familiar with untranslated units like *GB*,  *MB* and so on.  OSX's Finder is an example.

 
storage_units

This PR leaves these units untranslated except single letter *B* for *byte* (バイト).

EDIT: minor English grammar


kuroda:
@sakuro Thanks a lot.

I am in favor of this proposal, but I want to wait a few days to see responses from users.

PR #673, opened 2016-10-17
Add datetime.distance_in_words.x_hours

davetchen:
Add x_hours to be able to use it in the time distance helpers. The locales we use are

en*, ja, zh, ko, es, pt-BR

### DIFF
--- a/rails/locale/en-AU.yml
+++ b/rails/locale/en-AU.yml
@@ -87,6 +87,9 @@ en-AU:
       x_minutes:
         one: 1 minute
         other: "%{count} minutes"
+      x_hours:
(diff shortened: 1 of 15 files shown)


sreuterle:
Is that even useful when it's not supported by `DateHelper#distance_of_time_in_words`? I'd support changing both parts :)

kuroda:
Thank you for suggestions, but we cannot accpet your pull request.

Here we are collecting the translation data which have the same structure as that provided by the Rails itself.

See https://github.com/rails/rails/blob/master/actionview/lib/action_view/locale/en.yml

Please make a pull request to the repository of the Rails.

PR #674, opened 2016-10-17
Add datetime.distance_in_words.x_hours (rails-4-x)

davetchen:
Add x_hours to be able to use it in the time distance helpers for Rails 4. The locales we use are

en*, ja, zh, ko, es, pt-BR


kuroda:
See #673

PR #695, opened 2016-12-18
Add missing translations in fa.yml

intuxicated:
* Add missing translation to fa.yml
* Add fa to completed locales
* replace all ' with "


kuroda:
@intuxicated 

I want to know how I should call this language before I merge your commit.

Farsi? Persian?

intuxicated:
@kuroda they are both valid.
It's officially called [Persian language](https://en.wikipedia.org/wiki/Persian_language), we also call it `Parsi`. `Farsi` is common among us and most people use `Farsi` in daily talks.

kuroda:
@intuxicated Thanks.

PR #697, opened 2016-12-27
update ne.yml

abiralsambhu:
"गरिनुपर्नेछ" changed to " गरियो" in line 109 as it is the most relevant translation for the word "accepted".

### DIFF
--- a/rails/locale/ne.yml
+++ b/rails/locale/ne.yml
@@ -106,7 +106,7 @@ ne:
   errors:
     format: "%{attribute} %{message}"
     messages:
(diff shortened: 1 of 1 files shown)


shivabhusal:
@abiralsambhu thee context is "Active Record's data validation" so, in that context your changes do not make sense. It should be as is.

abiralsambhu:
@shivabhusal Previously I was unaware of that context. I thought of actual translation to that particular word **Accepted** to ** गरियो** . Thank You Dai :)

PR #705, opened 2017-01-28
Add x_years translation for ja.

treby:
I imported `en` and `ja` translation to my rails project, and noticed an missing key. So added it.

before

```bash
bundle exec rake i18n-spec:completeness rails/locale/en.yml rails/locale/ja.yml
### rails/locale/ja.yml

- *MISSING* datetime.distance_in_words.x_years
```

after

```bash
bundle exec rake i18n-spec:completeness rails/locale/en.yml rails/locale/ja.yml
### rails/locale/ja.yml

- *COMPLETE* 
```

Thank you for useful template:)

### DIFF
--- a/rails/locale/ja.yml
+++ b/rails/locale/ja.yml
@@ -93,6 +93,9 @@ ja:
       x_months:
         one: 1ヶ月
         other: "%{count}ヶ月"
+      x_years:
+        one: 1年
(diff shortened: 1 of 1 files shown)


kuroda:
Thanks a lot!

ありがとうございます。

PR #707, opened 2017-02-04
Updated Korean translation

ryush00:


keepcosmos:
What is purpose of this change? 
This change makes errors. Also it ignore convension.

@ryush00 @kuroda

PR #722, opened 2017-03-11
Correct long format in localized ES date

awernick:
In spanish, a year is referenced as "del" not "de".
Ex. "23 de Octubre del 2017"

### DIFF
--- a/rails/locale/es.yml
+++ b/rails/locale/es.yml
@@ -40,7 +40,7 @@ es:
     - sábado
     formats:
       default: "%d/%m/%Y"
-      long: "%d de %B de %Y"
+      long: "%d de %B del %Y"
       short: "%d de %b"
     month_names:
     -


Knack:
Sorry, @awernick but you are wrong

kuroda:
@awernick Do you have any counterargument?

awernick:
@Knack One of your sources proves me right.
http://www.fundeu.es/escribireninternet/como-se-escriben-las-fechas/
Picture

Knack:
Thanks @awernick, now I see that both are accepted, especially since the year 2000. For previous years, the correct form is "de" (e.g. "de 1999").

Anyway, I think we should keep "de" because it's valid for any date.

If you _know_ it's different in a specific region, like the US, you could submit a PR for that file (e.g. `es-US.yml`) but I wouldn't change it in the default `es.yml`

awernick:
@Knack totally agree. My primary spanish dialect comes from Mexico, so we mostly use "del" instead of "de". If I find enough sources to support the change to "de" to "del" in my specific region, then I'll submit a PR request.

PR #728, opened 2017-05-26
add 12 hour format to locale files

NehaGautam:
Add 12 hour format in our locale files, as many countries use 12 hour format example  US, UK, PH, CA (https://travel.stackexchange.com/questions/34950/which-large-countries-use-12-hour-time-format-am-pm). 
Instead of providing full format whenever calling I18n.l method we can then easily use 12 hour format. If you like this idea, I can make changes in all other locale files.


kuroda:
Thank you for suggestions, but we cannot accpet your pull request.

Here we are collecting the translation data which have the same structure as that provided by the Rails itself.

See https://github.com/rails/rails/blob/master/activesupport/lib/active_support/locale/en.yml

PR #734, opened 2017-07-30
Updated pt-BR.yml file with Rails 5.1.2. Added and translated default…

juliuscarvalhus:
Updated pt-BR.yml file with Rails 5.1.2. Added and translated default text en.yml to pt-BR. Applied standards according en.yml
:)


juliuscarvalhus:
Hi! :) 
About "pb:" and "eb:" in "number.human.storage_units" ... (lines 109 and 110)
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/locale/en.yml
About "hello" for test... (line 33)
https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/config/locales/en.yml

thank you so much! :)

PR #737, opened 2017-08-07
Add missing datetime translations

larouxn:
According to the README, all of the locales changed here are already considered complete. However, I figured I'd suggest a few missing translations, based off the existing ones. With these missing translations added, the changed locales now all have the exact same keys within `datetime`, key parity. Note, I'm perfectly fine with closing it if we're not interested/don't think it's a good idea.

EDIT: As I reverted my removal of two French translations, this is no longer a "synchronize translations" PR. Rather, I'm simply proposing we add a few missing translations and remove some trailing space from `en.yml`.

### DIFF
--- a/rails/locale/fr.yml
+++ b/rails/locale/fr.yml
@@ -95,6 +95,9 @@ fr:
       x_months:
         one: 1 mois
         other: "%{count} mois"
+      x_years:
+        one: un an
+        other: "%{count} ans"
       x_seconds:
         one: 1 seconde
         other: "%{count} secondes"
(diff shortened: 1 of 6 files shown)


rails/locale/fr.yml
```diff
@@ -76,11 +76,9 @@ fr:
         other: presque %{count} ans
       half_a_minute: une demi-minute
       less_than_x_minutes:
-        zero: moins d'une minute
```
larouxn: I didn't think about this before but some people might be using these `zero` translations. I need to revert this change and add the equivalent `zero` for the other locales.

PR #738, opened 2017-08-13
Occitan version

Quent-in:
Based on the General Occitan, also called "Standard Occitan" or "Broad Occitan".
It uses the mainly used forms of a same word to be easily understood by speakers of other dialect, eg different from the Lengadocian one.
Occitan is spoken en South France, North Spain, Monaco and North Italy.

### DIFF
--- a/rails/locale/oc.yml
+++ b/rails/locale/oc.yml
@@ -15,7 +15,7 @@ date:
     - dc
     - dj
     - dv
-    - ds
+    - dc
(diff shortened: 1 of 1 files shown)


kuroda:
@Quent-in 

I will check why the CI build is failing. Please wait for a while.

Quent-in:
I think I found what was wrong, indentation.
Let's see if it works now :)

PR #746, opened 2017-09-03
Create kab.yml

mozillakab:
Kabyle language file for rails-i18n

### DIFF
--- a/rails/locale/kab.yml
+++ b/rails/locale/kab.yml
@@ -1,0 +1,215 @@
+---
+fr:
+  activerecord:
+    errors:
+      messages:
+        record_invalid: 'Asentem ur yeddi aya : %{errors}'
+        restrict_dependent_destroy:
+          has_one: "Ur izmir ara ad yekkes akalas acku  %{record} seg-sen  yeneṭeḍ"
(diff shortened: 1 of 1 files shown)


BoFFire:
Need some check. It didn't pass the check.

rails/locale/kab.yml
```diff
@@ -0,0 +1,215 @@
+---
+fr:
+  activerecord:
+    errors:
+      messages:
+        record_invalid: 'Asentem ur yeddi aya : %{errors}'
```
BoFFire: "Asentem ur yeddi aya : %{errors}"
guillemets

rails/locale/kab.yml
```diff
@@ -0,0 +1,215 @@
+---
+fr:
```
kuroda: This line must be `kab:`.

digitalfrost:
Fails with

```

  1) I18n backend reloading when called twice doesn't produce warnings
     Failure/Error: errors.should == ''
     
       expected: ""
            got: nil (using ==)
     # ./spec/integration/backend_reload_spec.rb:21:in `block (3 levels) in <top (required)>'
```

PR #750, opened 2017-10-01
Create kab.yml

mozillakab:
Add kab locale

### DIFF
--- a/rails/locale/kab.yml
+++ b/rails/locale/kab.yml
@@ -1,0 +1,213 @@
+---
+kab:
+  activerecord:
+    errors:
+      messages:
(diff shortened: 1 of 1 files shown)


digitalfrost:
Fails with:

```
1) I18n backend reloading when called twice doesn't produce warnings
     Failure/Error: errors.should == ''
     
       expected: ""
            got: nil (using ==)
     # ./spec/integration/backend_reload_spec.rb:21:in `block (3 levels) in <top (required)>'
Finished in 0.72332 seconds (files took 0.33679 seconds to load)
1 example, 1 failure
```

This is the same as the problem with #746

PR #762, opened 2018-01-06
rename Belarusian locale file Fixes #760

### LINKED ISSUE #760: Belarus translation (BY) incorrectly saved as Belgium (BE)
I was looking for translations for Belgium and noticed that the be.yml file contains Belarusian translations. I believe be.yml should be changed to by.yml.

digitalfrost:


digitalfrost:
**by** is not correct 
#760 is wrong
Source: https://www.science.co.il/language/Locale-codes.php
This also matches the iso gem: https://github.com/tigrish/iso/blob/master/locales/en.yml
and is why the test fails.

PR #763, opened 2018-01-31
Include 'million' pluralisation option for fr and fr-FR

eleanorakh:
We're maintaining an app with several languages, and our French team have suggested that the default behaviour for the plural of `million` to be `millions`. 
This proposed change follows the examples provided in `rails/locale/es.yml`.

### DIFF
--- a/rails/locale/fr-FR.yml
+++ b/rails/locale/fr-FR.yml
@@ -171,7 +171,9 @@ fr-FR:
         format: "%n %u"
         units:
           billion: milliard
-          million: million
+          million:
+            one: million
+            other: millions
           quadrillion: million de milliards
           thousand: millier
           trillion: billion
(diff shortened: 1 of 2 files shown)


DevDef:
Why not achieving the work? The same apply to milliers (except if you replace it by “mille” which is invariable), milliard, millions de milliards, billions et trillions.
```
        format: "%n %u"
        units:
          billion:
            one: milliard
            other: milliards
          million:
            one: million
            other: millions
          quadrillion:
            one: million de milliards
            other: million de milliards
          thousand:
            one: millier
            other: milliers
          trillion:
            one: billion
            other: billions
```

Moreover, the plural in French is for absolute values of two and above, not different than one like in English. A good localization should respect these rules…
For example:
- 0 thing
- 1 thing
- 1.99 thing
- 2 things
- 1.9 million things
- 2 millions things
- -1.99 thing
- -2 things
- -1.9 million things
- -2 millions things

PR #776, opened 2018-04-06
Italian: Changing some string about validation

FiloSpaTeam:
I'm changing those strings to have more realistic and "gender-indipendent" forms.
For example:
* *Before*: **Biografia non puo** essere lasciato in bianco
* *After*: Biografia: **il campo non puo** essere lasciato in bianco

### DIFF
--- a/rails/locale/it.yml
+++ b/rails/locale/it.yml
@@ -109,35 +109,35 @@ it:
   errors:
     format: "%{attribute} %{message}"
     messages:
-      accepted: deve essere accettata
-      blank: non può essere lasciato in bianco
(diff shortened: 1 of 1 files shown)


kuroda:
Thanks a lot. But, I can't merge your commit immediately.

Because locale data of other languages do not include `:` at the head of translations, your commit could have an impact on the application developers.

I want to see if your commit is welcomed or not by the users of the `rails-i18n`.

pama:
We really appreciate your taking the time to submit this change, but we didn't have feedback for three years.

PR #781, opened 2018-04-28
Correcting grammatical error in Danish file

jesperronn:


jesperronn:
In case you wonder, one year is 'et år' in Danish. And not 'en år'. This was already correct other places in the file.

PR #788, opened 2018-07-24
Improve french locales

Silex:
Hello,

Everything is pretty much self-explanatory, ask if it is unclear.

~~Normalization makes it much easier to compare locales, because all the keys are alphabetically sorted. For example `fr.yml` and `fr-CA.yml` has keys at different positions, making it a bit of a pain to diff in the editor.~~ See #791

French is my mother tongue, it makes no sense to lowercase the days/months/million/etc names, also some of the translations are just plain incorrect, for example

``` 
Vous ne pouvez pas supprimer l'enregistrement car une personne à charge %{record} existe
```

Translates to:

```
You can not delete the record because someone dependant %{record} exists
```

After some tests, the translation came straight from google translate.

### DIFF
--- a/rails/locale/fr-CA.yml
+++ b/rails/locale/fr-CA.yml
@@ -64,40 +64,40 @@ fr-CA:
     distance_in_words:
       about_x_hours:
         one: environ une heure
-        other: environ %{count} heures
+        other: environ %{count} heures
       about_x_months:
         one: environ un mois
-        other: environ %{count} mois
+        other: environ %{count} mois
       about_x_years:
         one: environ un an
-        other: environ %{count} ans
+        other: environ %{count} ans
       almost_x_years:
         one: presqu'un an
         other: presque %{count} ans
       half_a_minute: une demi-minute
       less_than_x_minutes:
-        one: moins d'une minute
-        other: moins de %{count} minutes
-        zero: moins d'une minute
+        one: moins d'une minute
+        other: moins de %{count} minutes
+        zero: moins d'une minute
       less_than_x_seconds:
-        one: moins d'une seconde
-        other: moins de %{count} secondes
+        one: moins d'une seconde
+        other: moins de %{count} secondes
         zero: moins d'une seconde
       over_x_years:
         one: plus d'un an
-        other: plus de %{count} ans
+        other: plus de %{count} ans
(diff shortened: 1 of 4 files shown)


michaelbaudino:
Hrm, from what I understand, there is actually no reason to upcase week days and month names :thinking:

They are nouns and thus shall be written in lowercase (according to the [Académie Française](http://www.academie-francaise.fr/questions-de-langue#42_strong-em-jours-de-la-semaine-pluriel-et-majuscules-em-strong)).

I don't think "octet" (as a unit not derived from a proper name) should be upcased neither, fwiw.

Silex:
Right. I just discovered that in english weekdays are capitalized. I thought there were not. I'll remove the capitalization changes.

Silex:
Okay the commits are ready for the changes we should all agree about.

Now the following questions remain:

* Should we downcase https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml#L189 ?
* Should we downcase https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml#L174 ?

michaelbaudino:
I must admit I would not capitalize `Byte` neither, but I can't tell for sure (I'm not a native English speaker).

In French, for units, I would use uppercase when they are derived from a proper name and lowercase when they are not (_e.g._ `mètre`, `gramme`, `octet` but `Volt`, `Ampère`, `Newton`, etc...).
In English, no idea ¯\\\_(ツ)\_/¯

Silex:
Alright :+1: 

@svenfuchs, @kuroda: how should we proceed? merge this already and then open a new issue about the capitalization of `Byte` and `Million` in english?

kuroda:
To be honest, I don't like to see an ordering like "few", "many", "one", "other", "two", "zero". They are ordered alphabetically but unnaturally.

At least I want to wait until your proposal gathers some upvotes.

Please make another PR for the fixation of french locales.

Silex:
@kuroda: the normalization PR is at #791. I'll update this PR to be only about the french corrections.

Silex:
Ok I updated the commits for this PR to be only about the french modifications. The commits should be self-explanatory.

Silex:
:+1:

PR #791, opened 2018-08-06
Normalize locales

Silex:
Hello,

Normalization ensures the quoting is the simplest, that the line break width is respected and last but not leasts it sorts all the keys alphabetically, making it much easier to compare locales. For example `fr.yml` and `fr-CA.yml` have keys at different positions, making it a pain to diff in the editor.

**EDIT**: the following disadvantages are not true anymore, see below

The only disadvantage is when counts or units are involved:

``` yaml
      about_x_hours:
        few: ປະມານ %{count} ຊົ່ວໂມງ
        many: ປະມານ %{count} ຊົ່ວໂມງ
        one: ປະມານ 1 ຊົ່ວໂມງ
        other: ປະມານ %{count} ຊົ່ວໂມງ
        two: ປະມານ %{count} ຊົ່ວໂມງ
        zero: ປະມານ %{count} ຊົ່ວໂມງ
```

In this case it's probably more convenient to have it like this:

``` yaml
      about_x_hours:
        zero: ປະມານ %{count} ຊົ່ວໂມງ
        one: ປະມານ 1 ຊົ່ວໂມງ
        two: ປະມານ %{count} ຊົ່ວໂມງ
        few: ປະມານ %{count} ຊົ່ວໂມງ
        many: ປະມານ %{count} ຊົ່ວໂມງ
        other: ປະມານ %{count} ຊົ່ວໂມງ
```

I think the disadvantage is minor compared to the advantages.

### DIFF
--- a/rails/test/lib/normalize.rb
+++ b/rails/test/lib/normalize.rb
@@ -1,0 +1,43 @@
+require 'active_support'
+
+# Inspired from https://stackoverflow.com/questions/7275952/how-can-i-sort-yaml-files/13368706#13368706
+
+module Normalize
+  PLURALIZATION_KEYS = %w(zero one two few many other).freeze
+  ACTIVERECORD_KEYS = %w(has_one has_many).freeze
+  DATETIME_KEYS = %w(second minute hour day month year).freeze
+
+  class << self
+    def deep_sort(object)
+      return object unless object.is_a?(Hash)
+
+      hash = RUBY_VERSION >= '1.9' ? Hash.new : ActiveSupport::OrderedHash.new
+      object.each { |k, v| hash[k] = deep_sort(v) }
+
+      keys = hash.keys.map(&:to_s)
+      sorted = hash.sort do |a, b|
+        a_key = a[0].to_s
+        b_key = b[0].to_s
+        if (keys - PLURALIZATION_KEYS).empty?
+          PLURALIZATION_KEYS.index(a_key) <=> PLURALIZATION_KEYS.index(b_key)
+        elsif (keys - ACTIVERECORD_KEYS).empty?
+          ACTIVERECORD_KEYS.index(a_key) <=> ACTIVERECORD_KEYS.index(b_key)
+        elsif (keys - DATETIME_KEYS).empty?
+          DATETIME_KEYS.index(a_key) <=> DATETIME_KEYS.index(b_key)
+        else
+          a_matches = a_key.match(/^(.+)(#{DATETIME_KEYS.join('|')})(.*)$/)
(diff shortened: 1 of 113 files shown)


Silex:
Also, some locales are already alphabetically sorted, for example:

https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/fr-CA.yml#L81

Some commits also refered to this idea of sorting keys: efa75f9

kuroda:
@Silex Thanks.

I will merge this in near future unless there is any strong objection.

kuroda:
@Silex Please resolve conflicts.

Silex:
@kuroda: done

kuroda:
@Silex Could you add some instructions to the "How to contribute" section of the `README.md`.

Silex:
@kuroda: done, I also prettified the existing code blocks.

Silex:
Friendly ping :-)

rails/test/lib/normalize.rb
```diff
@@ -0,0 +1,15 @@
...
+
+# Taken from https://stackoverflow.com/questions/7275952/how-can-i-sort-yaml-files/13368706#13368706
+
+module Normalize
+  class << self
+    def deep_sort(object)
+      return object unless object.is_a?(Hash)
+      hash = RUBY_VERSION >= '1.9' ? Hash.new : ActiveSupport::OrderedHash.new
+      object.each { |k, v| hash[k] = deep_sort(v) }
+      sorted = hash.sort { |a, b| a[0].to_s <=> b[0].to_s }
```
schuetzm: Can you check for the keys `zero`, `one`, `two`, `few`, `many`, `other` and make sure they always appear before all other key and in that order, but only if there are no other keys present? I would much prefer to keep the logical order for pluralization rules.

e.g.
``` ruby
PLURALIZATION_KEYS = %w( zero one two few many other ).freeze
# ...
      sorted = if (hash.keys.map(&:to_s) - PLURALIZATION_KEYS).empty?
        hash.sort { |a, b| PLURALIZATION_KEYS.index(a[0].to_s) <=> PLURALIZATION_KEYS.index(b[0].to_s) }
      else
        hash.sort { |a, b| a[0].to_s <=> b[0].to_s }
      end
```

Silex: I think it's reasonable to have an exception in that case. Will do!

schuetzm:
Good idea to sort the other keys as well, this looks great!

Silex:
Alright, @kuroda, @digitalfrost :tada:

Silex:
Rebased to master.

@kuroda: can you comment a bit what your plan about this PR is?

Silex:
@kuroda: friendly ping

Silex:
Thanks! :+1:

PR #805, opened 2018-10-25
uppercase months and days for es-co

mariovzc:


digitalfrost:
Is this capitalization of months and days specific to Colombian Spanish? All the other Spanish locales seem to have lowercase months and days...

I think the rule might be the same as in French where days of the week are common names and thus are not capitalized unless they are at the beginning of a sentence....

jlduran:
@kuroda: Please revert this commit.

Source: http://www.rae.es/consultas/mayuscula-o-minuscula-en-los-meses-los-dias-de-la-semana-y-las-estaciones-del-ano

PR #817, opened 2018-12-17
Use full month names in abbr_month_names for arabic

mshka:
Abbreviations/acronyms are not used for the Arabic language 👉 https://www.quora.com/Why-arent-acronyms-used-in-Arabic/answer/Michael-Moszczynski 

_( there are very few cases where abbreviations exist but it is also barely used https://en.wiktionary.org/wiki/Category:Arabic_abbreviations )_


mshka:
@kuroda can you please have a 👀 at this

PR #818, opened 2018-12-17
Change first letter to lowercase in month names (LV).

pirminis:
Commit https://github.com/svenfuchs/rails-i18n/commit/8afe760daea7a18afc2ba7429964314b6c5dee09#diff-3ed775dd41617dc89c55676c782b7304 changed case of first letter of each of month names.

This is incorrect according to https://meta.wikimedia.org/wiki/Capitalization_of_Wiktionary_pages#Capitalization_of_month_names (see **Languages that do not capitalize month names:**)

**EDIT**
I asked native latvian translators on what correct translations in this case are, and currently in this PR the values are all correct.


pirminis:
this is NOT the final fix, it seems there is a bigger problem

PR #837, opened 2019-03-29
Revert "Bump i18n"

KaanOzkan:
Reverts svenfuchs/rails-i18n#835

Config change is actually wrong. Changing the application config will not change `I18n.locale`. Seems like `I18n.locale` is set from the config only once.

Reverting my changes until I confirm the best way to achieve this. Right now it seems to be setting `I18n.locale = nil` after initializing.

https://github.com/rails/rails/blob/master/activesupport/lib/active_support/i18n_railtie.rb#L18


KaanOzkan:
@kuroda time for a review?

PR #848, opened 2019-05-24
Singularize all datetime prompts

Silex:
Hello,

These changes looks fairly logical, but I understand it raises backward compatibility issues.


Silex:
:+1:

PR #854, opened 2019-06-20
Lower case month and day names in Portuguese language files

mynock:
Month and day names should not be capitalized by default in Portuguese. 

See https://meta.wikimedia.org/wiki/Capitalization_of_Wiktionary_pages#Capitalization_of_month_names and https://blogs.transparent.com/portuguese/months-in-portuguese/ for instance.


mynock:
@kuroda Any thoughts on what it would take to get this merged? Thanks!

PR #858, opened 2019-08-08
Fix negative currency format

frenkel:
Improve negative currency format based on https://taaladvies.net/taal/advies/vraag/1198/euro_12_12/

### DIFF
--- a/rails/locale/nl.yml
+++ b/rails/locale/nl.yml
@@ -156,6 +156,7 @@ nl:
       format:
         delimiter: "."
         format: "%u %n"
+        negative_format: '%u -%n'
         precision: 2
         separator: ","
         significant: false


frenkel:
Apparently I shouldn't fix it in this gem? Does anybody know where this is orignally defined wrong for the Dutch language?

pama:
@frenkel, I don't see that label on ActiveSupport [en.yml](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/locale/en.yml).

I'm closing this PR but feel free to continue the discussion if you feel I'm wrong. Thanks.

frenkel:
@pama where would this be defined then? It's incorrect for Dutch and I would like to fix it.

pama:
@frenkel from:

https://github.com/rails/rails/blob/ee95658b258005047c89cc64b1ee56823becfda1/activesupport/lib/active_support/locale/en.yml#L55-L67

It seems that is not predicted.

If you set that attribute in your locale, will it work with `number_to_currency` without sending it as an option? I didn't try that nor looked at the Rails source code.

frenkel:
Yes, when I place it in my Rails app translations it just works.

pama:
Oh, I see (I found [this](https://github.com/rails/rails/blob/984c3ef2775781d47efa9f541ce570daa2434a80/activesupport/lib/active_support/number_helper/number_to_currency_converter.rb#L41). In that case, I suggest you submit a change to the Ruby on Rails repository, perhaps add it to the "# These six are to override number.format and are optional" section.

I can't reopen the PR, as it seems that the pull request was deleted. I'm sorry.

frenkel:
I'm not sure I understand you correctly. What would I request to change in a PR to the Ruby on Rails repository?

Regarding your last statement: does it mean you would consider merging my change if I resubmit a PR to you?

pama:
Is my understanding that the [en.yml](https://github.com/rails/rails/blob/ee95658b258005047c89cc64b1ee56823becfda1/activesupport/lib/active_support/locale/en.yml) file should look like:

``` yaml
# in en.yml

# Used in NumberHelper.number_to_currency()
currency:
  format:
    # Where is the currency sign? %u is the currency unit, %n is the number (default: $5.00)
    format: "%u%n
    unit: "$"
    # These seven are to override number.format and are optional     <===============================
    separator: "."
    delimiter: ","
    precision: 2
    # number_to_currency:   <==============================
    # round_mode:
    significant: false
    strip_insignificant_zeros: false
```

Since we use these yaml files as references, I suggest to open a PR in [rails/rails](https://github.com/rails/rails) to add the missing translation. It seems a lapse.

I would consider keeping your PR open until we see the change in the RoR side. I can't tell if we will merge it before the change. I would rather wait, but other collaborators may have a different opinion.

PR #859, opened 2019-08-10
Polish: Update pl.yml

meceo:
more common language.

### DIFF
--- a/rails/locale/pl.yml
+++ b/rails/locale/pl.yml
@@ -3,10 +3,10 @@ pl:
   activerecord:
     errors:
       messages:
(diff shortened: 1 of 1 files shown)


digitalfrost:
Looks good to me. 

But could someone else that speaks Polish please confirm that this new phrasing is better?

@meceo - could you give some more details ... is the phrasing you propose less formal... or do you mean that the new phrasing is more typical of what Polish websites would use in this context...

marcinwierzbicki:
This polish translation is a little bit less formal than current. I think new one is better suited for websites.

maro-21:
I confirm the translation is correct and the pull request can be merged.

PR #869, opened 2019-09-03
Automatically eager load I18n if possible

casperisfine:
Since `i18n 1.6.0` you can call `I18n.eager_load!` so that all translations are loaded immediately during the boot phase, rather than of the first request.

So ideally if we detect this, we add it automatically to `eager_load_namespaces` so that users don't have to know about it.

cc @rafaelfranca @Edouard-Chin


ghiculescu:
@casperisfine is this needed, given that the behaviour is also included in Rails since https://github.com/rails/rails/pull/37756 ?

casperisfine:
Yeah, fine by me.

ghiculescu:
https://github.com/svenfuchs/rails-i18n/pull/983

PR #878, opened 2019-11-27
Update locale for vi (vietnamese)

imcvampire:


imcvampire:
> @imcvampire the key `taken` seems to already exist on line 127, which one is correct?

Both are good in Vietnamese. I will create a PR to remove one

imcvampire:
https://github.com/svenfuchs/rails-i18n/pull/902

PR #892, opened 2020-03-03
Add pluralization rule for pt-BR locale

maykonmenezes:
I've noticed that pluralization rule for pt-BR is missing. This PR adds it

### DIFF
--- a/rails/pluralization/pt-BR.rb
+++ b/rails/pluralization/pt-BR.rb
@@ -1,0 +1,3 @@
+require 'rails_i18n/common_pluralizations/one_other'
+
+::RailsI18n::Pluralization::OneOther.with_locale(:'pt-BR')


Throne3d:
According to [this Unicode supplemental](https://www.unicode.org/cldr/charts/29/supplemental/language_plural_rules.html), (Brazilian) Portuguese uses "one" for 0 and 1, but "other" for everything else (including decimals like 0.1). That's the `OneWithZeroOther` pluralization, not `OneOther`, right? (European Portuguese does follow `OneOther`, though.)

I'm also new to this project, but to check – it looks like the `pt` tag is being used for European Portuguese, right? What standard does the project follow for language tags? I'm asking because it seems like Unicode assigns `pt` to Brazilian Portuguese, needing to specify `pt_PT` for the European variant, and this seems to match what Chrome's JavaScript `Intl` implementation does, at least guessing by the pluralization rules (not sure what the spec is for that):

```js
[0, 0.1, 1, 2].map(x => new Intl.PluralRules('pt-PT').select(x)) // ["other", "other", "one", "other"]
[0, 0.1, 1, 2].map(x => new Intl.PluralRules('pt').select(x))    // ["one", "one", "one", "other"]
[0, 0.1, 1, 2].map(x => new Intl.PluralRules('pt-BR').select(x)) // ["one", "one", "one", "other"]
```

pama:
@Throne3d pt is the translation to European Portuguese.

PR #895, opened 2020-03-18
Finnish: Improvements to error and datetime prompt translations

teemukonkola:
Minor improvements to Finnish translations. Rationales behind the changes:
- `prompts`: These seem to be uppercase in other languages, so doing the same here for consistency.
- `blank`: The old term translates literally to "contentless", which sounds a bit unnatural. Better to use "empty" instead.
- `invalid`: The old term translates to something like "unacceptable". The new one could be "invalid", "incorrect" or "erroneous"
- `not_an_integer`: Changing from "is not an integer" to "must be an integer", which is consistent with the English locale.
- `one`: Changing from "error prohibited..." to "error in input prohibited...", more descriptive.

### DIFF
--- a/rails/locale/fi.yml
+++ b/rails/locale/fi.yml
@@ -17,7 +17,7 @@ fi:
     - pe
     - la
     abbr_month_names:
-    - 
+    -
     - tammi
     - helmi
     - maalis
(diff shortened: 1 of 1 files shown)


sergioisidoro:
Hi there @kuroda 

We've had these changes in our project for a while, so we've decided propose them here. 
Tell us if you need any additional information. 

PS: Teemu is a native Finnish speaker, and these strings were also reviewed by @emilvirkki (also native speaker)

PR #896, opened 2020-04-04
Brazilian Portuguese: Translate 'short' to curta instead of curto

LucasKuhn:
The 'too_short' error is mainly used when showing errors about password, in which case it would be better translated as 'curta' ( A senha )

### DIFF
--- a/rails/locale/pt-BR.yml
+++ b/rails/locale/pt-BR.yml
@@ -134,8 +134,8 @@ pt-BR:
         one: 'é muito longo (máximo: 1 caracter)'
         other: 'é muito longo (máximo: %{count} caracteres)'
       too_short:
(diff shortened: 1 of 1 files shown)


sergioisidoro:
Hi @LucasKuhn 

Have you checked this?
https://github.com/siefca/i18n-inflector-rails

I don't think there's native support for gender in rails internationalisation, but since Portuguese generally defaults to masculine, and this is general key, maybe keeping it masculine makes more sense?

pama:
I'm sorry @LucasKuhn, but we will not merge this PR unless we get massive support from the Portuguese community.

We can not assume that too_short will only be used with passwords (Palava-passe / Senha), and we default to masculine in all Portuguese variations (pt-PT and pt-BR).

Issue #890, opened 2020-02-20
Columbian Spanish should use the decimal comma


alekseyg:
A client brought up an issue with currency and numbers being formatted incorrectly in our app when using the es-CO locale. After looking it up, I've confirmed that Columbia uses the comma as the decimal separator and the full stop as the thousands separator. However, the es-CO locale in rails-i18n uses the US format of full stop decimal separator and comma thousands separator.

https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma

digitalfrost:
Closed by #898

PR #898, opened 2020-05-19
Colombian Spanish (es-CO): Change number and currency delimiters

ldeld:
See https://github.com/svenfuchs/rails-i18n/issues/890 

The es-CO locale was incorrectly using the English-speaking style of number delimitation, fixed to follow the common Spanish speaking style (https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma)


jlduran:
Close #890.

PR #899, opened 2020-05-23
Add x_hours to de.yml

kaelumania:
I had an issue with the current rails version, that a translation ist missung for datetime distance in words `x_hours`

### DIFF
--- a/rails/locale/de.yml
+++ b/rails/locale/de.yml
@@ -91,6 +91,9 @@ de:
       x_minutes:
         one: eine Minute
         other: "%{count} Minuten"
+      x_hours:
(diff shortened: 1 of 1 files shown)


kuroda:
Unfortunately, this PR can not be accepted.

This is because the `datetime.distance_in_hours.x_hours` key does not exist in the Rails YAML file. Have a look at this.

https://github.com/rails/rails/blob/master/actionview/lib/action_view/locale/en.yml

It's strange that this key is missing, but we just provide translations of the YAML files provided by Rails.

Please consider sending a PR to https://github.com/rails/rails.

Issue #897, opened 2020-04-25
Correct dates in Polish


maro-21:
Hello,
I came here from [OpenStretMap](https://github.com/openstreetmap/openstreetmap-website/issues/2602) because they use your script.

The dates on OpenStreetMap website don't display correctly.

We use four different date formats but two of them are incorrect:
- "kwiecień 25, 2020" (on the user's profile page)
- "25 kwiecień 2020" (messages and other places)

"kwiecień 25, 2020" looks like an American date format filled in with Polish words, but in Polish we never start a date with a month.

There is a list of months
https://github.com/svenfuchs/rails-i18n/blob/26f1bd90c2cd9e7192fbf84d64a5415e6a2fc977/rails/locale/pl.yml#L45-L58

It is indeed a list of months but... in the nominative case and for dates we use the genitive case, so if these names of months are used only for dates (I suppose so), they should be changed to:
```
    - stycznia
    - lutego
    - marca
    - kwietnia
    - maja
    - czerwca
    - lipca
    - sierpnia
    - września
    - października
    - listopada
    - grudnia
```

Please correct also:
https://github.com/svenfuchs/rails-i18n/blob/26f1bd90c2cd9e7192fbf84d64a5415e6a2fc977/rails/locale/pl.yml#L25
to "maja".

So the correct format for the dates mentioned above is:
25 kwietnia 2020

Issue #729, opened 2017-07-06
pl: long date format is invalid


mspanc:
Currently long date format for the pl locale is specified as follows:

 long: "%B %d, %Y"

It produces e.g.

`lipiec 06, 2017 20:11`

it is not natural for polish native speakers.

The right is

6 lipiec 2017, 20:11

(day without leading 0 / month / full year / comma / datetime)

but it is gramatically incorrect, as there are cases in polish and month names should not be in their primary form. Here's the list:

stycznia
lutego
marca
kwietnia
maja
czerwca
lipca
sierpnia
września
października
listopada
grudnia

but that form should be used ONLY in the long format, for rest of the cases month names that are currently present in the locale should remain.

kuroda:
@mspanc Thank you for reporting.

It would be great if you submit a pull request. I will merge it as soon as possible.

adas172002:
Hi,

I would like to point to your attention that there is a gem [polish](https://github.com/grk/polish) that tries to address this special requirement of Polish language. Perhaps a solution applied in this gem can be also used for rails-i18n.

digitalfrost:
Closed by #903

PR #903, opened 2020-07-26
Use Polish long date format (with day number at the begging)

wiktorn:
Use Polish long date format (with day number at the begging)

I haven't change the month names to genitive case as I'm not sure if this would be OK in all cases.

Closes: #897


digitalfrost:
Also closes #729

PR #910, opened 2020-10-01
Revert apostrophes that are less known that right quotes

benoittgt:
True apostrophes are the best way of writing in french but this is not well
known and rarely used outside of printed documents. French rarely knows
how to type it, and rarely use them in webpages.

Partially revert: https://github.com/svenfuchs/rails-i18n/pull/889

Related: https://github.com/svenfuchs/rails-i18n/pull/889#issuecomment-701964424

### DIFF
--- a/rails/locale/fr.yml
+++ b/rails/locale/fr.yml
@@ -5,9 +5,9 @@ fr:
       messages:
         record_invalid: 'La validation a échoué : %{errors}'
         restrict_dependent_destroy:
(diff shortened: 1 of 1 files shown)


rouaneta (approved):
Thanks @benoittgt, let's be pragmatic :)

benoittgt:
Hello @kuroda 

Do you think it is possible to merge?

PR #911, opened 2020-10-02
Add Ruby 2.7.1 + bump previous ruby

benoittgt:
While working on other PR. I noticed that Ruby versions in CI are not the last one, and that Ruby 2.7 is not present.

See: https://www.ruby-lang.org/en/downloads/releases/


benoittgt:
CI is green now.

PR #922, opened 2021-03-02
Translate ordinals

sunny:
Hello everybody! 👋🏻 

Since Rails 6 accepts [translations for ordinalize](https://github.com/rails/rails/pull/32168), we might as well add them to rails-i18n as well!

This lets apps do:

```rb
I18n.locale = :en
1.ordinalize # => "1st"

I18n.locale = :fr
1.ordinalize # => "1er"
```

### DIFF
--- a/rails/ordinals/fr-CA.rb
+++ b/rails/ordinals/fr-CA.rb
@@ -1,0 +1,19 @@
+{
+  :"fr-CA": {
+    number: {
+      nth: {
+        ordinals: -> (_key, number:, **_options) {
+          if number.to_i.abs == 1
+            'er'
+          else
+            'e'
+          end
+        },
+
+        ordinalized:  -> (_key, number:, **_options) {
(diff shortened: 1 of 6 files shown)


gagalago (approved):

sunny:
Thanks for the review @gagalago and thanks for maintaining rails-i18n @kuroda! ✨✨

rails/ordinals/fr-CA.rb
```diff
@@ -0,0 +1,19 @@
+{
+  :"fr-CA": {
```
casperisfine, line 2: 👋 I'm getting a syntax error on this, not sure why CI didn't catch it:

```ruby
>> {:"fr-CA": nil}
`eval': (irb):1: syntax error, unexpected ':', expecting => (SyntaxError)

sunny: Oh darn. Tests cover only the `fr.rb` file. I’ll add a failing test for all locales.

PR #936, opened 2021-07-03
ES: Standardize precision for number_to_human

### LINKED ISSUE #928: I believe this is not correct
https://github.com/svenfuchs/rails-i18n/blob/99cfa35268ad20ec1193c670a7ee2a4705140a3d/rails/locale/es.yml#L186

In english this presicion is 3, in spanish is 1

So having

number_to_human(15, precision: 1)

gives 20, which is misleading.

number_to_human(1511, precision: 1)

gives "2 mil" ("2 Thousand") which is way far from 1,51 Thousand

I'll adjust this in my own locales but I suggest putting it in 3, same as english

jlduran:
Close #928

### DIFF
--- a/rails/locale/es-CL.yml
+++ b/rails/locale/es-CL.yml
@@ -183,7 +183,7 @@ es-CL:
           unit: ''
       format:
         delimiter: ''
-        precision: 1
+        precision: 3
         significant: true
         strip_insignificant_zeros: true
       storage_units:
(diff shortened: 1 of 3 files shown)


jlduran:
For Spanish, the rules are a bit more complex (see https://www.rae.es/dpd/números). However, having a `precision: 3` for all territorial variants is recommended.

PR #939, opened 2021-08-06
Fix multiple locale date formats per CLDR

### LINKED ISSUE #938: Many locales' long date formats are incorrect
It seems like what originally might've been the locale file for the `en-US` locale specifically has been copied for many of the other locales, which have incorrect date formats (`Month DD, YYYY` is only used in the US and some parts of Canada). These should all be fixed as appropriate.

getsnoopy:
Fixes #938

### DIFF
--- a/rails/locale/af.yml
+++ b/rails/locale/af.yml
@@ -40,8 +40,8 @@ af:
     - Saterdag
     formats:
       default: "%Y-%m-%d"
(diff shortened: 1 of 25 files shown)


MoskitoHero:
OK, I will take the time to review this PR and make sure we don't introduce any typo.

If anyone can give it a second look too, you are welcome to do so.

PR #941, opened 2021-08-15
There is no space between attribute and message in Traditional Chinese

khiav223577:
Correct: "名稱不能為空白"
Incorrect: "名稱 不能為空白"

---

Take [Microsoft docs](https://docs.microsoft.com/zh-tw/troubleshoot/mem/intune/troubleshoot-windows-enrollment-errors ) for example, it use:

- 名稱不能為所有數位。
- 名稱不能包含空格。

### DIFF
--- a/rails/locale/zh-TW.yml
+++ b/rails/locale/zh-TW.yml
@@ -107,7 +107,7 @@ zh-TW:
       month: 月
       year: 年
   errors:
-    format: "%{attribute} %{message}"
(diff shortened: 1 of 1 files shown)


sunny:
@khiav223577 should this change also be applied to [zh-CN](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/zh-CN.yml)?

PR #955, opened 2021-12-13
allow usage of rails 7

dorianmariefr:

PR #957, opened 2021-12-16
Make rails-i18n work with rails7

### LINKED ISSUE #954: railties dependency and rails 7 compatibility
First, thanks to everyone who works on this gem.

I recently ran into a compatibility issue when running `bundle update` against rails/main (currently at https://github.com/rails/rails/commit/a35a380c2cdc7fe032f927095545aee21a192f54):

```ruby
Bundler could not find compatible versions for gem "railties":
  In Gemfile:
    rails was resolved to 7.1.0.alpha, which depends on
      railties (= 7.1.0.alpha)

    rails-i18n (~> 6.0.0) was resolved to 6.0.0, which depends on
      railties (>= 6.0.0, < 7)
```

Noting that not much has changed between rails-7.0.0rc1 (which bundled fine, obviously) and rails/main, I forked this repo and (naively) changed the dependency in rails-i18n.gemspec from

`s.add_runtime_dependency('railties', '>= 6.0.0', '< 7')`

to

`s.add_runtime_dependency('railties', '>= 6.0.0')`

Everything bundled fine and I didn't notice anything weird with my translations. Obviously I didn't do any extensive testing with the gem itself.

Anyway, in brief, I was curious if there were any known compatibility issues with railties >= 7.0 that would warrant keeping the dependency above or whether a version bump would be appropriate.

Thanks again.

f6p:
EDIT: Temp Quickfix
-------------------
In your **Gemfile** replace `gem 'rails-i18n'` with `gem 'rails-i18n', git: 'https://github.com/f6p/rails-i18n.git', branch: 'rails7-hotfix'`. See my [comment below](https://github.com/svenfuchs/rails-i18n/pull/957#issuecomment-999021980).

Original PR message
-------------------

Rails7 was released today. *gemspec* has hardcoded depndency `rails < 7`. As a consequence *bundler* can resolve dependencies on release candidates but not on stable release.

PR is similar to #955 but does two more things. It updates `Gemfile.lock` (I ran specs - one translation seems to be missing) and fixes invalid hash syntax in `rails/ordinals` (that caused errors when testing application with *rails-i18n* dependency).

It fixes #954, replaces: #955.

Cheers
🍷


Gemfile.lock
```diff
@@ -48,62 +47,62 @@ GEM
...
+    terminal-table (3.0.2)
+      unicode-display_width (>= 1.1.1, < 3)
+    thor (1.1.0)
+    tzinfo (2.0.4)
+      concurrent-ruby (~> 1.0)
+    unicode-display_width (2.1.0)
+    zeitwerk (2.5.1)
 
 PLATFORMS
-  ruby
```
dorianmariefr, line 106: this should probably stay "ruby"

rails-i18n.gemspec
```diff
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = '>= 1.8.11'
 
   s.add_runtime_dependency('i18n', '>= 0.7', '< 2')
-  s.add_runtime_dependency('railties', '>= 6.0.0', '< 7')
+  s.add_runtime_dependency('railties', '>= 7.0.0', '< 8')
```
uxxman, line 19: Since everything still works with rails 6, I think we should only relax the upper and not restrict the lower one i.e.

```ruby
s.add_runtime_dependency('railties', '>= 6.0.0', '< 8')
```

f6p:
I removed ordinal changes as now there is a better PR for that (#958).

rails-i18n.gemspec
ryanb: Agreed, supporting both Rails 6 and 7 will help with transitioning.

morgoth: @f6p can you please do this small change?

dorianmariefr: I think it should just be `>= 6` so that it's compatible with future versions of rails in an "evergreen" setup

f6p:
Guys as a temporary quickfix we are using this in our Gemfile:
```
gem 'rails-i18n', git: 'https://github.com/f6p/rails-i18n.git', branch: 'rails7-hotfix'
```
Its my branch that merges #958 and this one (#957) and works well for us.

No need to wait for official release (you can switch to gem at later time).

morgoth:
@f6p sure, but can you please do this small change in this PR? https://github.com/svenfuchs/rails-i18n/pull/957#discussion_r772333098

davidwessman:
I used @f6p commit and added the proposed change to support Rails 6 but opened in #962.
I kept @f6p as author as well.

MoskitoHero:
Closing this in favour of #962

PR #968, opened 2022-01-04
update taken error for fr translation

cprodhomme:
the translation for `taken` was the same of `exclusion`

But in the usage (for exemple when your email is already taken), the sentence "n'est pas disponible" is not really clear


MoskitoHero:
I agree that I have always found the previous translation quite unclear. I believe we can merge this unless anyone comes up with an objection.

Issue #965, opened 2021-12-28
Siberian pluralization warning


ryanb:
Since upgrading to rails-i18n 7.0.1, I see the following warnings when calling `I18n::JS.filtered_translations`.

```
$ rails c
>> I18n::JS.filtered_translations
/Users/rbates/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/rails-i18n-7.0.1/lib/rails_i18n/../../rails/pluralization/sr.rb:4: warning: already initialized constant I18n::Backend::Base::RailsI18n::Pluralization::Serbian::FROM_2_TO_4
/Users/rbates/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/rails-i18n-7.0.1/lib/rails_i18n/../../rails/pluralization/sr.rb:4: warning: previous definition of FROM_2_TO_4 was here
/Users/rbates/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/rails-i18n-7.0.1/lib/rails_i18n/../../rails/pluralization/sr.rb:5: warning: already initialized constant I18n::Backend::Base::RailsI18n::Pluralization::Serbian::FROM_12_TO_14
/Users/rbates/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/rails-i18n-7.0.1/lib/rails_i18n/../../rails/pluralization/sr.rb:5: warning: previous definition of FROM_12_TO_14 was here
```

I assume this is from PR #925. However I'm uncertain why the pluralization files are loaded multiple times. Is that an issue with [i18n-js](https://github.com/fnando/i18n-js)? If so I can post this there instead.

Is anyone else seeing this?

pama:
@ryanb I created a new rails 7 app, with `gem 'rails-i18n', '~> 7.0.1'` and `gem "i18n-js"`, using ruby `3.0.2p107`. I set locale to `:sr`, and I couldn't reproduce that warning on my rails console. Let me know if I need to do something else to see that warning.

ryanb:
@pama Thanks for looking into this.

It sounds like there's something specific in my app causing the double loading. I'll close this for now and will update when/if I find the culprit.

nickcampbell18:
@ryanb I'm seeing the same thing, did you figure out what this was in the end?

ryanb:
@nickcampbell18 I haven't had a chance to look into this yet, but it's on my todo list. 😃

[...]

PR #972, opened 2022-01-05
Add Serbian as a common pluralizations file

dougedey-shopify:
Fixes https://github.com/svenfuchs/rails-i18n/issues/965

This caused some test failures on our codebase so I wanted to fix it upstream.

It appears that the pattern for all the other specialized translations is in `common_pluralizations` with the full name of the language, so I replicated that behaviour and the issue appears to have been fixed (at least locally)

I think I moved the files to the right locations, tests for the serbian locale still pass with this change (I did force them to fail locally)

### DIFF
--- a/lib/rails_i18n/common_pluralizations/serbian.rb
+++ b/lib/rails_i18n/common_pluralizations/serbian.rb
@@ -1,0 +1,32 @@
+module RailsI18n
+  module Pluralization
+    module Serbian
+      FROM_2_TO_4   = (2..4).to_a.freeze
+      FROM_12_TO_14 = (12..14).to_a.freeze
+
+      def self.rule
+        lambda do |n|
+          n ||= 0
+          mod10 = n % 10
+          mod100 = n % 100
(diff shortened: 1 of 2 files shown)


lib/rails_i18n/common_pluralizations/serbian.rb
```diff
@@ -0,0 +1,33 @@
...
+            { locale => {
+                :i18n => {
+                  :plural => {
+                    :keys => [:one, :few, :other],
+                    :rule => rule }}}}
+        end
+      end
+    end
+  end
+
```
shioyama, line 33:
```suggestion
```

dougedey-shopify: Whoops, also fixed the indentation

toncid:
The same change applies to Croatian, Bosnian and Serbo-Croatian (politics aside 😄 ), so I have filed ticket #986 to cover all of them properly.

PR #982, opened 2022-02-14
Add yalphabetize to ci

samrjenkins:
Multiple locale files in the repo have not been normalized. Adding the yalphabetize gem to our rake tasks and CI will allow us to catch locale yaml files that require normalisation before merging a PR.

This PR adds yalphabetize to our default rake task and corrects any un-alphabetised locale files.

Yalphabetize also offers a tool to alphabetise any un-alphabetised yaml files. In the future, we could potentially use this to replace the rails-i18n Normalize module.

### DIFF
--- a/.yalphabetize.yml
+++ b/.yalphabetize.yml
@@ -1,0 +1,23 @@
+allowed_orders:
+  - [has_one, has_many]
+  - [zero, one, two, few, many, other]
+  - [second, minute, hour, day, month, year]
+  -
+    - about_x_hours
+    - about_x_months
+    - about_x_years
+    - almost_x_years
+    - half_a_minute
+    - less_than_x_seconds
+    - less_than_x_minutes
+    - over_x_years
+    - x_seconds
+    - x_minutes
+    - x_days
+    - x_months
+    - x_years
+
+
+indent_sequences: false
+only:
+  - rails/locale/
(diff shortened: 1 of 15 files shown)


MoskitoHero:
Nice addition. I agree that skimming through unsorted YAML files is a pain. The only issue I see here is adding a new gemfile dependency, which I am not against per se, but I think we need the opinion of several people before we merge that kind of stuff.

So, is there any objection from the community?

@samrjenkins  A recent merge has updated the gemfile, which now conflicts with your version. Could you update it ?

MoskitoHero:
> Will take a look at CI failures shortly

The CI was recently moved to Github actions and fails due to [a bug in the CY translation](https://github.com/svenfuchs/rails-i18n/issues/990). This behavior is unrelated to your MR.

digitalfrost:
Link to yalphabetize  on [rubygems](https://rubygems.org/gems/yalphabetize) for more information on the gem.
The github repo  can be found [here](https://github.com/samrjenkins/yalphabetize)

digitalfrost:
Please look at #791 which tried to resolve this issue.
There is a task to normalize the locals in the thor file https://github.com/svenfuchs/rails-i18n/blob/master/locales.thor
normalize sorts all the keys alphabetically...
**yalphabetize** seems like a nice gem - but would it's addition be better than the current solution?

pama:
@samrjenkins I do appreciate your effort, but I'm with @digitalfrost here. For now, we will try to address this another way.

Issue #881, opened 2019-12-10
currency shouldn't use a simple space, use nbsp instead.


estani:
### Summary 
The problem was already mentioned in  [apidock in 2008](https://apidock.com/rails/ActionView/Helpers/NumberHelper/number_to_currency#259-number-to-euro).
There is no valid scenario where a symbol should be separated from the amount it describes, it's as bad as breaking the number itself, so this behavior should be finally addressed, imho.

### Steps to reproduce
Use some locale where the currency symbol has a space in it, like `de`
```ruby
HTMLEntities.new.encode(Class.new.extend(ActionView::Helpers::NumberHelper).number_to_currency(12, locale: 'de'), :named)
=> "12,00 &euro;"
```
Because of the simple space the semantic is:
> if needed is ok to separate "12,00" from "&euro;"

which is not true.

### Expected behavior
The space need to be considered part of the currency, so it should be a non-breaking space.
The expected result should be: `12,00&nbsp;&euro;`

### Work around
Replacing the space by `nbsp` would sort it out (a proven work around for over 11 years):
```ruby
HTMLEntities.new.encode(Class.new.extend(ActionView::Helpers::NumberHelper).number_to_currency(12, locale: 'de').gsub(' ',"\u00A0"), :named)
=> "12,00&nbsp;&euro;"
```
### Proposal

replace the `format: "%n %u"` of the locale (example of `de`:
https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/de.yml#L159)

by `format: "%n\u00A0%u"`

pama:
I don't understand why translations should address formatting. For specific apps, we can customize our own .yml file, or use a span with `white-space:nowrap;`

dvodvo:
Adding a span is possible, but still clutter. 
i18n does address formatting, but is a hack, and incomplete.  The seperator between number and unit is a universal concern for a site and should be consistent across all languages.

This is truly of the realm of the helper.

[based on my reading](https://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html#method-i-number_to_currency),  the number to currency method has `:format` attribute which should allow format: "%u[something for `&nbsp;` ]%n". This obviously has wider implications on all the helpers, for which I have insufficient perspective.

Creating a distinct attribute would create clutter, but may be the proper path to follow given the non-breaking space handling consequences.

ruurd:
[...]

PR #996, opened 2022-04-18
French (fr): Revert remainder of NBSP characters

johnnyshields:
In 6f21c25f89573eee05609b6ef1f1c1a12e9ac24d, @benoittgt _partially_ reverted "thin spaces" i.e. NBSP characters from French (`fr`) locale.

I do not understand the reason for a "partial" revert. It's not explained in the commit, and there seems to be no consistent logic regarding which NBSP chars were reverted. For example:

```
      too_long:
        other: est trop long (pas plus de %{count} caractères) # not reverted 🤷 
      too_short:
        other: est trop court (au moins %{count} caractères)   # reverted
```	

This PR reverts all NBSP chars from `fr`, and in the future if the maintainers would like to make some consistent rule on where/how NBSP should in which locales, let's re-apply it at that time.

### DIFF
--- a/rails/locale/fr.yml
+++ b/rails/locale/fr.yml
@@ -75,19 +75,19 @@ fr:
         other: environ %{count} ans
       almost_x_years:
         one: presqu'un an
-        other: presque %{count} ans
+        other: presque %{count} ans
       half_a_minute: une demi‑minute
       less_than_x_seconds:
         zero: moins d'une seconde
         one: moins d'une seconde
-        other: moins de %{count} secondes
+        other: moins de %{count} secondes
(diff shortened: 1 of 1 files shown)


pama:
That was reverted due to https://github.com/svenfuchs/rails-i18n/pull/889#issuecomment-701912770 I believe.

In my opinion, rails-i18n should not address nbsp, at all (for some reasons here #881 and here #889)

pama:
@DevDef @arkadiybutermanov @jbbarth @benoittgt

benoittgt (approved):

PR #1001, opened 2022-06-09
Adds ordinal translation for spanish, portuguese and german.

rafafloresta:
For spanish, adds a º to right of the number.
For portuguese, adds a º to right of the number.
For german, adds a . to right of the number.

Also added `# frozen_string_literal: true` to existing files, as it is done on the rails codebase to convert the keys to symbols.

### DIFF
--- a/rails/ordinals/de-AT.rb
+++ b/rails/ordinals/de-AT.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+{
+  :"de-AT" => {
+    :ordinal => proc { |number| "#{number}." }
+  }
+}
--- a/rails/ordinals/es-419.rb
+++ b/rails/ordinals/es-419.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+{
+  "es-419": {
+    number: {
+      nth: {
+        ordinals: lambda do |_key, number:, **_options|
+          abs_number = number.to_i.abs
+          "º"
+        end,
+
+        ordinalized: lambda do |_key, number:, **_options|
+          "#{number}#{ActiveSupport::Inflector.ordinal(number)}"
+        end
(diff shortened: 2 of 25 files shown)


rails/ordinals/es-419.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+{
+  "es-419": {
+    number: {
+      nth: {
+        ordinals: lambda do |_key, number:, **_options|
```
pama, line 6: :number is not a keyword argument: https://github.com/rails/rails/commit/1e4e0935bcf14cd5ab7c384ea9e3b91c4a8387b7

spec/unit/ordinals_spec.rb
```diff
@@ -34,4 +34,42 @@
...
+          ActiveSupport::Inflector.ordinalize(2).should == "2º"
+          ActiveSupport::Inflector.ordinalize(3).should == "3º"
+        end
+      end
+    end
+  end
+
+  describe 'German' do
+    it 'uses the custom rules' do
+
```
pama, line 64: Could you remove the extra newlines?

pama (commented):
Hello,

Thank you for your pull request. Inline you can find two comments.

rails/ordinals/de-AT.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+{
+  :"de-AT" => {
+    :ordinal => proc { |number| "#{number}." }
```
sunny, line 4: The hashes for german can be harmonized to use use the shorter syntax, like for the other locales. E.g.:

```suggestion
  "de-AT": {
    ordinal: => proc { |number| "#{number}." }
```

Issue #577, opened 2015-09-07
thor locale:complete does not list complete locales


digitalfrost:
thor locale:complete does not list complete locales correctly
thor local:incomplete incorrectly lists complete locales as incomplete

Logically  en.yml which is the original from rails should always be complete.

```
gerald@gerald-VirtualBox:~/code/rails-i18n$ thor locales:complete

gerald@gerald-VirtualBox:~/code/rails-i18n$ thor locales:incomplete
af, ar, az, be, bg, bn, bs, ca, cs, cy, da, de, de-AT, de-CH, el, en, en-AU, en-CA, en-GB, en-IE, en-IN, en-NZ, en-US, en-ZA, eo, es, es-419, es-AR, es-CL, es-CO, es-CR, es-EC, es-MX, es-PA, es-PE, es-US, es-VE, et, eu, fa, fi, fr, fr-CA, fr-CH, gl, he, hi, hi-IN, hr, hu, id, is, it, it-CH, ja, km, kn, ko, lb, lo, lt, lv, mk, mn, mr-IN, ms, nb, ne, nl, nn, or, pl, pt, pt-BR, rm, ro, ru, sk, sl, sr, sv, sw, ta, th, tl, tr, tt, ug, uk, ur, uz, vi, wo, zh-CN, zh-HK, zh-TW, zh-YUE
gerald@gerald-VirtualBox:~/code/rails-i18n$ i18n-tasks missing en
Missing translations (0) | i18n-tasks v0.3.9
Good job! No translations missing!
gerald@gerald-VirtualBox:~/code/rails-i18n$ i18n-tasks missing es
Missing translations (0) | i18n-tasks v0.3.9
Good job! No translations missing!
gerald@gerald-VirtualBox:~/code/rails-i18n$ 
```

This is the same bug as #546 - hopefully this ticket makes the problem clearer

EddieOne:
Is it broken? I have this issue as well, I think. https://github.com/carmen-ruby/carmen-rails/issues/13#issuecomment-518155799

jlduran:
Fixed by #1087.

PR #1010, opened 2022-08-12
Add method to count incomplete locales

digitalfrost:
Added method to count incomplete locales to locales.thor
Refactored incomplete to make this easier to add.


digitalfrost:
Helps with #577 and #1011

Issue #1013, opened 2022-08-14
thor list and count_all ignore iso-639-2 locales


digitalfrost:
The locales in the iso-639-2 sub-directory are not returned by
`thor locales:list` 
or counted by 
`thor locales:count_all`

Also see #1011 [Problem Counting Locales](#1011)

digitalfrost:
The iso-639-2 locales are also not listed in the _Available Locales_ section of the README

PR #1015, opened 2022-08-15
README.md: Add missing locales to list of available locales

digitalfrost:
Include locales that are in the iso-639-2 sub-directory


digitalfrost:
Fixes #1013

Issue #1011, opened 2022-08-12
Problem Counting Locales


digitalfrost:
The README lists 109 **available locales**.
Counting the incomplete locales returned by ```thor locales:incomplete```
using #1010 returns 116 ***incomplete locales***.

See also #577

digitalfrost:
```
$ tree rails/locale
...
1 directory, 122 files
```
Indicating we have 122 locale.yml files

digitalfrost:
```
$ ls -F rails/locale |grep -v / | wc -l
116
```

digitalfrost:
Using #1012
```
$thor locales:count_all
116
```

digitalfrost:
```
$tree rails/pluralization
0 directories, 151 files
```

digitalfrost:
We have locales like yo (Yoruba) that have pluralization rules but no locale file.

digitalfrost:
The orphaned pluralizations are not all the same.
For example the pluralization rules specified for yo (Yoruba) and wa (Walloon) are not the same.

They however come from the same commit [392c65f](https://github.com/svenfuchs/rails-i18n/commit/392c65fb1f086d1b5708e4be0301d500deb884cc) on Jan 29, 2012

PR #1016, opened 2022-08-16
Add method to list orphaned pluralizations

digitalfrost:
Add method to thor locales to list locales that
 have a pluralization file but no locale file
See https://github.com/svenfuchs/rails-i18n/issues/1011


pama:
Thanks! 🎉

PR #1017, opened 2022-08-17
Remove orphan pluralizations

digitalfrost:
- Deleted orphan pluralizations and corresponding specs.
- Added spec to test for orphan pluralizations.
- Enabled Rspec to use both _should_ and _expect_ syntax

#1011 #577


pama:
Thank you!

PR #1019, opened 2022-08-22
feat: module initialization options

aergonaut:
Allows the separate features of the gem to be selectively enabled. This allows an app to use only the pluralization rules, for example, and not the locale rules.

On my team, we have our own locale formatting rules that we want to keep using. However, we also want to add pluralization rules. We would like to use this gem to provide the pluralization rules, but we don't need the other parts of the gem. While we could manually add the appropriate files from this gem to the I18n load path, it would be nicer if the gem allowed us to configure which features we want to use.

The module names are just taken from the names of the directories passed to `add`. So `rails/locale` gets added if the `:locale` module is enabled.

Users would configure the gem in their `config/application.rb` like

```ruby
config.rails_i18n.enabled_modules = [:pluralization]
```

If they don't configure anything (the default) then the gem adds all features to the enabled list and functions as before. This ensures backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,31 @@ Note that your Ruby on Rails version must be 3.0 or higher in order to install t
 
 ## Configuration
 
+### Enabled modules
+
+By default, all `rails-i18n` modules (locales, pluralization, transliteration, ordinals) are enabled.
+
+If you would like to only enable specific modules, you can do so in your Rails configuration:
+
+```ruby
+# to enable only pluralization rules, but disable all other features
+config.rails_i18n.enabled_modules = [:pluralization]
+
+# to enable pluralization and ordinals
+config.rails_i18n.enabled_modules = [:pluralization, :ordinals]
+```
+
+The possible module names:
+
+* `:locale`
+* `:ordinals`
+* `:pluralization`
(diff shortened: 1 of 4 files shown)


aergonaut:
Fixed the failure in `spec/unit/ordinals_spec.rb` by providing `rails_i18n` to the `app.config` mock, and requiring `rails_i18n` instead of the railtie directly, so that the module config accessor would be present.

pama:
I think this is useful.

We need a reference to this in the README file.

aergonaut:
@pama I have added documentation to the README

pama:
Thanks. Let's wait to see if we get some feedback.

aergonaut:
@pama Do you want to wait more or could we merge this?

aergonaut:
@pama Could we merge this?

pama:
@aergonaut I'm still waiting for other maintainers' opinions about this. @digitalfrost @MoskitoHero?

MoskitoHero:
Yes I believe this is a good move and it should be merged.

aergonaut:
@pama @MoskitoHero Thank you!

PR #1020, opened 2022-08-26
Add method to list locales without pluralization files

digitalfrost:
Add methods to **check_locales** and to **thor locales** to list locales that
have a locale file but no pluralization file
See svenfuchs#1011


pama:
Thanks!

PR #1021, opened 2022-08-27
Add pluralization for some English language locales

digitalfrost:
Add pluralization for:
- en-US
- en-ZA
- en-TT
- en-IE
- en-CY


pama:
Thank you!

PR #1023, opened 2022-08-28
README: Document locales without pluralization

digitalfrost:
- added section to README to list the locales with missing pluralization
- removed **ad** from list of complete locales since it is missing a pluralization file


pama:
thanks!

PR #1025, opened 2022-08-29
Add transliteration rule for capital sharp s

padde:
The letter ẞ is the capital variant of the small letter ß used in German.
It's pretty rare in the wild, but here it is for completeness sake!

Also see https://en.wikipedia.org/wiki/%C3%9F#Development_of_a_capital_form


pama:
Thanks!

padde:
I just noticed an edge case for transliteration of capital letters in general. There is no handling of all-caps situations. Consider:

```
# capital sharp s only appears in all-caps so not a problem
I18n.transliterate("FUẞBALL")
#=> "FUSSBALL"

# umlauts currently handled only at beginning of a normally spelled noun correctly
I18n.transliterate("Ästhetik")
#=> "Aesthetik"

# however for all caps words it does produce a mixed case result:
I18n.transliterate("KANÜLE")
#=> "KANUeLE"
```

Of course there are more edge cases, like how do we handle single letter strings (I suggest they are basically also all caps, so should produce all caps transliterated).

Any thoughts on this? Should I dive into an implementation of this? Is this a general problem with all languages that needs more careful consideration?

pama:
Could you open an issue with these considerations?

PR #1029, opened 2022-08-31
Update gl.yml, "incluído" missing accent in gl.

torce:
- Incluido [participle without accent does not exist](https://academia.gal/dicionario/-/termo/busca/incluir), press "Conxugar" first form of "Participio" is "incluído".


pama:
Thanks!

PR #1031, opened 2022-09-02
Fixing some issue for Bengali

Aftabuzzaman:
I am Bengali wikipedia editor & admin https://bn.wikipedia.org/wiki/user:আফতাবুজ্জামান https://w.wiki/5ejH I hope i will qualify here for fixing some issue in Bengali.

I fixed some issue here:
- date & time - Bengali doesn't use "de" (I think spanish uses it). date & time should be like this: Friday, 2 September 2022
- fixed some spelling
- added translation for "byte"

### DIFF
--- a/rails/locale/bn.yml
+++ b/rails/locale/bn.yml
@@ -18,7 +18,7 @@ bn:
     - মে
     - জুন
     - জুলাই
-    - অগাস্ট
+    - আগস্ট
     - সেপ্টেম্বর
     - অক্টোবর
     - নভেম্বর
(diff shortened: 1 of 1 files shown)


digitalfrost:
Thank you.

Aftabuzzaman:
Thank for accepting.

I have a question: https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/bn.yml#L69 this line has opening & closing apostrophe (`'`), while English verison https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml#L79 doesn't have it. Will it cause any problem? or it doesn't matter?

I will submit another patch. `abbr_day_names:` & `abbr_month_names:` is not actually abbreviated here & same as `day_names:` & `month_names:`.

digitalfrost:
Looking at https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/bn.yml#L72
I don't think the apostrophes are needed.
There is nothing dynamic going on such as a `%{...}`
and there seems that there is nothing special about the opening or closing characters.
Well spotted!

PR #1039, opened 2022-09-08
Lao: simplify pluralization

digitalfrost:
No need for zero, one, two, few many, other

See #1033


johnnyshields:
👍 this looks good. I believe we can also get rid of the trailing spaces, but should check with a native speaker for that...

digitalfrost:
Weird - I don't have the trailing space in Vim.
The top part of the last character is a separate character...

digitalfrost:
I'll edit it via Github

PR #1049, opened 2022-10-06
Punjabi Gurmukhi minor corrections

bgo-eiu:
Punjabi Gurmukhi adjustments (Gurmukhi being the script used in India - I will submit a PR for Punjabi Shahmukhi `pnb` / `pa-PK`, the script used in Pakistan, soon)

Notes on the changes:

* Correction to order of attaching characters and diacritics, as well as addition of nukta 'dots' to words like ਜ਼ਰੂਰੀ where their typical spelling and pronunciation warrants their distinction. ਅੈ and ਬੱੁ render incorrectly in most fonts, including the fonts used on GitHub, so these are the most important fixes of this kind.

* The words for all the large numbers except 1000 were transliterations of the English, but there are Punjabi words for these numbers so I changed them to these. "Quadrillion" is particularly hard to pronounce in Punjabi as it involves multiple sounds which do not exist in the language.

* Used ਇੱਕ instead of 1 where applicable. This was already done in the first string, and I noticed this is done already in the Urdu locale as well (Urdu having the same word for one, just in a different script). The main reason for this is that multiple numeral systems may be used with Punjabi, so if the application allows the rendering of ੨ rather than 2, it would look strange for 1 to always only be 1.

* Translated the file size units to full words. This just makes things a little more cohesive; although the day/month abbreviations in this context are fine, abbreviations are only used sparingly in Punjabi if at all, and it is common for kilometre for example to be written out like this. (In the other script, Shahmukhi, abbreviations are almost never used for anything as they are not legible in that system.)

* Changed the last word separator to not use a comma. There is not a strict rule about this for Punjabi, but more often than not I do not really see much use of the Oxford comma like this. It is also less necessary within the context of how the grammar works compared to English - there would be no ambiguity in "the panda eats, shoots and leaves" vs "the panda eats, shoots, and leaves," for example. The word forms would have to change in order for the meaning to change like that, and in similar sentences the "and" may even be omitted or implied while remaining clear.

---

Relatedly, I have a question. I noticed the introduction of some ordinal format rules for French. Are these meant just for calender dates, or to be used with anything? I ask because Punjabi ordinals inflect for grammatical gender (and gender and case). However, the names of all the months are all masculine and singular if being used in a date, so if the ordinal formatter is only for dates, it could produce 6ਵਾਂ ਅਕਤੂਬਰ (6th of October). If it is generalizable to say, "the sixth cat," though then the format should be 6ਵੀਂ ਬਿੱਲੀ to agree with the common "cat" word being feminine, which complicates things. (Huge cats can be masculine but that's another story.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## unreleased
 
+- Punjabi Gurmukhi (pa): typographical corrections, native large number names, unit expansion, removal of Oxford comma #1049
 - French (fr, fr-CA, fr-CH, fr-CA): change an abreviation for March month in abbr_month_names #1002
 
 ## 7.0.5 (2022-07-03)
--- a/rails/locale/pa.yml
+++ b/rails/locale/pa.yml
@@ -9,10 +9,10 @@ pa:
           has_many: ਮਿਟਾ ਨਹੀਂ ਸਕਦੇ ਕਿਉਂਕਿ ਨਿਰਭਰ %{record} ਮੌਜੂਦ ਹਨ
   date:
     abbr_day_names:
-    - ਅੈਤ
+    - ਐਤ
     - ਸੋਮ
     - ਮੰਗਲ
-    - ਬੱੁਧ
+    - ਬੁੱਧ
     - ਵੀਰ
     - ਸ਼ੁੱਕਰ
     - ਸ਼ਨਿੱਚਰ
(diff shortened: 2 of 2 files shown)


sunny:
@bgo-eiu Awesome work! Thank you for providing the whole reasoning.

Could you update the CHANGELOG as well? 🙏🏻 

Regarding ordinal I’m afraid they are used for different use-cases, which is up to the application authors. In the Ruby on Rails documentation they are described as to be used for "[ordered sequences](https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-ordinalize)" with no option to specify a grammatical gender.

bgo-eiu:
Have now updated the changelog.

That's alright re: ordinals, was just curious. Someone really determined to use Punjabi ordinals like this would likely have to use a specific tool for it (or if this doesn't exist I might try making this)

pama:
@bgo-eiu could you resolve the conflicts? Thanks!

pama:
@bgo-eiu closing this due to lack of feedback. Feel free to open it again if you decide to fix the merge conflict. Thanks!

PR #1053, opened 2022-10-17
Added Punjabi Shahmukhi locale files

bgo-eiu:
These are for Punjabi in the Perso-Arabic Shahmukhi script as used in Pakistan. This is the update I mentioned in #1049 .

Due to inconsistency in how browsers and applications request this locale, both `pnb` and `pa-PK` are required. For example, on OpenStreetMap which has this project as a dependency, a user has to include both of these language codes in their locale settings as the login interface uses `pnb` but the editor uses `pa-PK`. (Depending on the platform for translations or the software, some combination of both codes, one or the other, or neither might be used.)


sunny:
> Depending on the platform for translations or the software, some combination of both codes, one or the other, or neither might be used.

Shouldn’t this choice be made at the application level instead? For example I think you can use `I18n.fallbacks[:pnb] = :"pa-PK"` in your `config/initializers/i18n.rb` for the same effect without having to duplicate all translation files.

bgo-eiu:
So I was not sure if this was typical or not given the presence of both `sv-SE` and `sv` for example.

I would like to set up fallbacks like this in the OpenStreetMap project, but apparently this is not trivial to do and they have various longstanding issues with it. There is a fallback for `no` to `nb` in their initializer config set up like this, but I tested it and it does not work.

sunny:
> There is a fallback for no to nb in their initializer config set up like this, but I tested it and it does not work.

Oh, yeah, the syntax seems really weird! It should be `I18n.fallbacks[:no] = :nb` instead of `I18n.fallbacks.map("no" => "nb")`.

pama:
@bgo-eiu could you please resolve the conflict?

pama:
@bgo-eiu closing this due to lack of feedback. Feel free to open it again if you decide to fix the merge conflict. Thanks!

PR #1055, opened 2022-10-18
Added sv-FI to locales

jolll4:
Added sv-FI to the list of locales. Finland has a large minority of native Swedish-speakers, who speak a slightly differing version of the language than those in Sweden. It can be benefitial to differentiate between the locales for platforms that operate in both countries.

The locale implemented has differing date format and currency from the sv locale.

### DIFF
--- a/rails/locale/sv-FI.yml
+++ b/rails/locale/sv-FI.yml
@@ -39,9 +39,9 @@ sv-FI:
       - fredag
       - lördag
     formats:
-      default: "%Y-%m-%d"
-      long: "%e %B %Y"
-      short: "%e %b"
+      default: "%-d.%-m.%Y"
+      long: "%d. %Bta %Y"
+      short: "%d. %b"
     month_names:
       -
       - januari
--- a/rails/locale/sv-FI.yml
+++ b/rails/locale/sv-FI.yml
@@ -1,0 +1,210 @@
+---
+sv-FI:
+  activerecord:
+    errors:
+      messages:
+        record_invalid: "Ett fel uppstod: %{errors}"
+        restrict_dependent_destroy:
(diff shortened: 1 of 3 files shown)


sunny:
Thank you! Could you also add a line to the CHANGELOG, under the unreleased section? 🙏🏻 

E.g.:

```md
- Add following locales:
  - Swedish (sv-FI): Finland’s native Swedish-speakers #1055
```

rails/locale/sv-FI.yml
```diff
@@ -0,0 +1,210 @@
...
+      x_minutes:
+        one: en minut
+        other: "%{count} minuter"
+      x_days:
+        one: en dag
+        other: "%{count} dagar"
+      x_months:
+        one: en månad
+        other: "%{count} månader"
+      x_years:
```
pama, line 99: I noticed that `x_years` is not present in https://github.com/rails/rails/blob/main/actionview/lib/action_view/locale/en.yml and is present in many of our translations. Is this a copy & paste issue, or I'm missing something?

sunny: Good find! It’s also not inside the actual helper, so this sounds like something to clean up in rails-i18n.  https://github.com/rails/rails/blob/3661d0d8a46a2725b3e7bacbc230e4f183fb67dc/actionview/lib/action_view/helpers/date_helper.rb#L154

jolll4: Should we change it for `sv-FI` here or separately along all the others in one go?

pama: @jolll4 could you compare your proposal to https://github.com/svenfuchs/rails-i18n/tree/master/rails/rails

Let's merge this locale correctly; another time, we will need to correct the remaining locale files. Thanks.

PR #1065, opened 2023-01-17
Transform to decimal to select pluralization rule

### LINKED ISSUE #1064: Passing string in French locale throws the following error "comparison of String with 0 failed"
There should be some error handling and type check that throws meaningful error message when passed argument is of a wrong type. 
Other locales like German or English do not throw error messages due to the Ruby handling equality comparisons ("1" == 1 -> false). As opposed to "2" > 1 -> "comparison of String with 1 failed (ArgumentError)"
This is inconsistent behavior
Judging from the unit tests, you only test "happy paths", there is no error cases testing.

sunny:
Closes #1064

There are several ways the issue could be solved:

- **Raise an error when given a non-numeric**
  This could break apps that don’t check that are passing strings.
  For example this would raise an `ArgumentError` when given `"1"`.

- **Consider all non-numerics as `other`**
  This is closer to the current `en` implementation however that is probably not what the user intended when passing numeric-looking strings.
  For example this would return `1 apples` when given `"1"`.

- **Try to be helpful and transform it to a numeric**
  For example this would return `1 apple` when given `"1"`.

I went for the latter. Let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## unreleased
 
-- Transform to decimal when checking pluralization rules #1065
+- Transform to decimal to select pluralization rule #1065
 - Update following locales:
   - Japanese (ja): Add `in` and `round_mode` keys #1059
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## unreleased
 
+- Transform to decimal when checking pluralization rules #1065
 - Update following locales:
   - Japanese (ja): Add `in` and `round_mode` keys #1059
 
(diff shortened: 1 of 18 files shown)


pama (approved):

sunny:
This covers the initial issue, so I think it can be merged 👍🏻

pama:
Thank you!

sunny:
And thank you for maintaining rails-i18n 🎉

rakris01:
@sunny @pama you need to update these file too.
https://github.com/svenfuchs/rails-i18n/blob/master/rails/pluralization/pl.rb
https://github.com/svenfuchs/rails-i18n/blob/master/rails/pluralization/ar.rb

PR #1066, opened 2023-01-18
Update currency formatting for South African Rand

izoomi:
Not a native speaker, but this would match how js formats it:

```
> new Intl.NumberFormat('en-ZA', { style: 'currency', currency: 'ZAR'}).format(10000)
> 'R 10 000,00'
```
Also, from a quick google search it seems this is the correct way (https://www.sadev.co.za/content/how-correctly-format-currency-south-africa):

> The so-called Continental System (also used in South Africa) requires that the decimal point be replaced by a comma
> ...
> Use a space, not commas, to indicate thousands

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Update following locales:
   - Japanese (ja): Add `in` and `round_mode` keys #1059
+  - English (en-ZA): ZAR currency format #1066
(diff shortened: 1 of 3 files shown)


pama:
@izoomi could you update the change log?

izoomi:
@pama updated

pama:
Thanks!

PR #1067, opened 2023-01-19
Non-numerics counts are considered as `other` in all pluralizations

sunny:
Following @Dmitri-Filatov’s suggestion on #1064, this changes any non-numeric counts to be considered as `other` in all pluralizations.

Compared to #1065: 

- This does not try to detect if the given string is numeric-looking or not, for performance and consistency between all pluralizations.

	```rb
	I18n.t("datetime.distance_in_words.x_days", count: "abc", locale: :en) # => "abc days"
	I18n.t("datetime.distance_in_words.x_days", count: "abc", locale: :fr) # => "abc jours"
	I18n.t("datetime.distance_in_words.x_days", count: "1", locale: :en) # => "1 days"
	I18n.t("datetime.distance_in_words.x_days", count: "1", locale: :fr) # => "1 jours"
	```

- For consistency, the `nil` value is always considered as `other` instead of sometimes being treated as `zero`, `one` or `other` depending on the locale.

- This includes and tests all locales (e.g. `ar`, `pl`), and not only the shared pluralizations (thanks @rakris01)

Although this does not handle string numbers `"1"` I think this behavior is actually simpler and more consistent. What do you think @pama?


Dmitri-Filatov:
@sunny Sorry, I was running IDE with old Ruby version accidentally (2.4). BigDecimal is fixed in later versions and will produce 0 instead of generating error.
The last comment I have is that 'abc'.to_d will generate 0. Maybe it should be "other" rather than zero?

sunny:
Done, it’s now considered as `other` 👍🏻

rakris01:
@sunny @pama  when are we merging this to master?

sunny:
For me, this is good to merge 🚀

pama:
>Although this does not handle string numbers "1" I think this behavior is actually simpler and more consistent. What do you think @pama?

@sunny I tested this using Rails without this gem, and it seems that numeric strings are handled as strings:

```
irb(main):012:0> I18n.t("datetime.distance_in_words.x_days", count: "1", locale:
 :en)
=> "1 days"
```

So we don't need to make this an exception.

pama:
Thank you for your work.

PR #1069, opened 2023-02-01
Add support for German ordinalization

kalsan:
This pull request adds German ordinalization. It's trivial (for once, German is actually simple (!)) but fixes the missing translation problem for German.

BEFORE:

```ruby
irb(main):001:0> 1.ordinalize
=> "translation missing: de.number.nth.ordinalized"
```

NOW: 

```ruby
irb(main):001:0> 1.ordinalize
=> "1."
```


sunny (approved):
Can you add an entry to the CHANGELOG ? 🙏🏻

PR #1075, opened 2023-04-03
Add gd locale

gunchleoc:
Let me know if I need to define the plural rules somewhere. They are described at https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html#gd

I have a question regarding `month_names`: Are these being used standalone as well as being part of `long: "%d %B %Y"`? If there is no standalone usage, I should change these before we merge.

Also, I don't have a ruby environment set up, so this is not tested. Will the CI do an adequate job to prevent any crashes?

### DIFF
--- a/rails/ordinals/gd.rb
+++ b/rails/ordinals/gd.rb
@@ -1,0 +1,23 @@
+{
+  gd: {
+    number: {
+      nth: {
+        ordinals: -> (_key, number:, **_options) {
+          if number.to_i.abs == 1
+            'ᵈ'
+          elsif number.to_i.abs == 2
+            'ⁿᵃ'
+          elsif number.to_i.abs == 3
+            'ˢ'
+          else
+            'ᵐʰ'
+          end
+        },
+
+        ordinalized:  -> (_key, number:, **_options) {
+          "#{number}#{ActiveSupport::Inflector.ordinal(number)}"
+        }
+      }
+    }
+  }
+}
--- a/rails/ordinals/gd.rb
+++ b/rails/ordinals/gd.rb
@@ -3,11 +3,12 @@
     number: {
       nth: {
         ordinals: -> (_key, number:, **_options) {
-          if number.to_i.abs == 1
+          case number.to_i.abs
+          when 1
             'ᵈ'
-          elsif number.to_i.abs == 2
+          when 2
             'ⁿᵃ'
-          elsif number.to_i.abs == 3
+          when 3
             'ˢ'
           else
             'ᵐʰ'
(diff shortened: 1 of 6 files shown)


sunny:
Hi there! Thank you for adding Scottish Gaelic! Could you also add an entry to the CHANGELOG.md file as well?

> I have a question regarding month_names: Are these being used standalone as well as being part of long: "%d %B %Y"? If there is no standalone usage, I should change these before we merge.

Yes apps can use these as they like, without requiring them to use the `long` format.

For example:

```rb
> I18n.l(1.month.ago, format: "%B", locale: :en)
=> "March"
> I18n.l(1.month.ago, format: "%B", locale: :fr)
=> "mars"
```

I don’t know what the specific change you had in mind for Scottish Gaelic, but for French, as you can see the month name is usually lowercase. This can sometimes be an issue if authors using I18n assume it is always capitalized.

gunchleoc:
Thanks for the info!

I have added tests to the spec file and also included a rule for ordinals.

The change I had in mind would be changing "1 Am Foilleach" to "1ᵈ dhen Fhoilleach" for "1st of January" etc, but it's better to keep the standalone from when they can't be handled separately.

rails/ordinals/gd.rb
```diff
@@ -0,0 +1,23 @@
...
+        ordinals: -> (_key, number:, **_options) {
+          if number.to_i.abs == 1
+            'ᵈ'
+          elsif number.to_i.abs == 2
+            'ⁿᵃ'
+          elsif number.to_i.abs == 3
+            'ˢ'
+          else
+            'ᵐʰ'
+          end
```
sunny, line 14: A minor change in syntax here can help reduce repetition and is slightly more performant:

```suggestion
          case number.to_i.abs
          when 1
            'ᵈ'
          when 2
            'ⁿᵃ'
          when 3
            'ˢ'
          else
            'ᵐʰ'
          end
```

gunchleoc:
@sunny Thanks, I have committed your suggestion

sunny (approved):

pama:
Tests are failing.

rails/pluralization/gd.rb
```diff
@@ -0,0 +1,29 @@
+module RailsI18n
+  module Pluralization
+    module ScottishGaelic
+      def self.rule
+        lambda do |n|
+          return :other unless n.is_a?(Numeric)
+
+          floorn = n.floor()
```
sunny, line 8: In Ruby we’d usually drop the parentheses here.

```suggestion
          floorn = n.floor
```

sunny (approved):

sunny:
> Will this be handled if I add a "zero" key there

For which translation key would you want to handle the zero?

sunny:
Looks good to me 👍🏻

gunchleoc:
Thank you all for your help & review. My first Rails code!

PR #1080, opened 2023-05-10
Remove duplicate keys

KaikuOMantere:
When parsing, noticed some keys which are present twice. Looks accidental


sunny (approved):
Awesome! Could you add a small entry to the CHANGELOG?

sunny (approved):

PR #993, opened 2022-04-11
Change `1` in `one` key to `%{count}`

movermeyer:
### What are you trying to accomplish?

Fixes #823. cc @cejaekl

`one` is used for numbers other than 1.

> A common mistake is to think that "one" is only for only the number 1.
> Instead, "one" is a category for any number that behaves like 1.
> So in some languages, for example, one → numbers that end in "1" (like 1, 21, 151) but that don't end in 11 (like 11, 111, 10311).
- http://cldr.unicode.org/index/cldr-spec/plural-rules

### What approach did you choose and why?

Simple find/replace, then looked through the changes to make sure they were correct.

Find: `one:(.*)1(.*)`
Replace: `one:$1%{count}$2`
Then `git add -up`

### What should reviewers focus on?

🤷 

### The impact of these changes

Correct translations when lookups are done in locales that use `one` for non-1 counts.

PR #1086, opened 2023-06-04
Update rails update script

jlduran:
The `update.rb` script downloads the locale files from [rails/rails](https://github.com/rails/rails).
These files are merged and normalized into a single `en.yml`, which serves as the basis for this gem.
This pull request updates the `update.rb` script, by pointing it to the correct branch from the Rails project. Also, updates the latest version of each locale file under `rails/rails`. It is important to note that these files are external to this gem, and only the keys are used for comparison.


rails/rails/action_view.yml
```diff
@@ -4,37 +4,37 @@
     distance_in_words:
       half_a_minute: "half a minute"
       less_than_x_seconds:
-        one:   "less than %{count} second"
+        one:   "less than 1 second"
```
sunny, line 7: The decision to prefer `%{count}` was discussed in https://github.com/svenfuchs/rails-i18n/pull/993

jlduran: Correct, that is why #1088 leaves them as they are.
These changes are verbatim from https://github.com/rails/rails/blob/main/actionview/lib/action_view/locale/en.yml
This file does not belong to this project, it is only used as a reference.

jlduran:
I have updated the description to further clarify the changes, and state that the files we are changing are under `rails/rails` i.e., these files are external files, not under the control of this project. Only the keys of these files are used for comparison.

sunny (approved):

pama (approved):

PR #1090, opened 2023-06-04
es: Change caracter -> carácter

jlduran:
The word _caracter_ (without accent) does not exist.
It should be spelled _carácter_ regardless of region.
The plural form _caracteres_ is correct.

Source: https://www.rae.es/dpd/carácter


sunny (approved):
👍🏻 Can you add an entry in the CHANGELOG ?

jlduran:
Thank you! CHANGELOG entry added.

sunny (approved):

PR #1091, opened 2023-06-14
Fix Rails' update script branch

jlduran:
We should track the latest Rails release, not the main branch.

Fix 30a8745 by pointing the update script to the proper Rails branch.

Perhaps changing to a git model that more closely mimics Rails' is a better approach long-term.  In the meantime, keep bumping the branch in the update script as Rails releases are published.

Fix #1086.


jlduran:
@pama: This is an alternative approach: We should align to the latest production Rails release.

PR #1093, opened 2023-06-20
Use hryvnia symbol as a currency unit in uk.yml

NARKOZ:


sunny (approved):
Thanks! Could you also add an entry to the CHANGELOG?

NARKOZ:
Added

PR #1104, opened 2023-10-17
Make short Portuguese date format consistent in length with other languages

tferrerm:
Both Portuguese and Brazilian Portuguese have abbreviated month names. It makes more sense to use them when printing a date in the short format, as is done in most other languages (including Spanish, where 'de' is also used in the short date format).

### DIFF
--- a/rails/locale/pt-BR.yml
+++ b/rails/locale/pt-BR.yml
@@ -41,7 +41,7 @@ pt-BR:
     formats:
       default: "%d/%m/%Y"
       long: "%d de %B de %Y"
-      short: "%d de %B"
+      short: "%d de %b"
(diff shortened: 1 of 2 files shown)


pama:
@tferrerm, thank you for your contribution. However, I am not inclined to accept this request. As a native speaker of European Portuguese, the "12 de jan" format is not a commonly used abbreviation. The standard and preferred form is "12 de janeiro."

Abbreviations such as "12/jan/2023" or "12-jan-2023" can be used, but using numbers is the most commonly preferred format.

tferrerm:
@pama what about `%d %b`? This would output `12 jan` for example.

PR #1105, opened 2023-10-17
Add indication of AM or PM to all locales where missing

tferrerm:
When specifically formatting a timestamp in the 12-hour clock format, it should be possible to distinguish between AM and PM, even if the 12-hour clock is not commonly used in said langauge. Also, "AM" and "PM" are pretty much universal and most people around the world should understand that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 7.0.9 (2023-11-16)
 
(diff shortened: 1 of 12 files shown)


pama:
@tferrerm could you please update the change log?

sunny (approved):

pama:
@tferrerm I'm sorry, but now we have a small conflict that needs a fix. Could you fix it for us?

PR #1108, opened 2023-11-01
[da] Do not use title case

c960657:
Danish does not use Title Case for in headlines and definitely not in expressions like “128 Bytes” or “5,4 Millioner”. Apart from actual movie/book titles, Title Case is almost only used with proper nouns, certain titles and pronouns and in abbreviations.

The authoritative rules for letter case in Danish are described here:
https://sproget.dk/raad-og-regler/Retskrivningsregler/retskrivningsregler/a7-11-14


pama:
@c960657 I noticed a change from KB to kB, but the other units are uppercase.

rails/locale/da.yml
```diff
@@ -194,11 +194,12 @@ da:
         format: "%n %u"
         units:
           byte:
-            one: Byte
-            other: Bytes
+            one: byte
+            other: bytes
           gb: GB
-          kb: KB
+          kb: kB
```
sunny, line 200:
```suggestion
          kb: KB
```

sunny (commented):
Can you also add a CHANGELOG entry? 🙏🏻

c960657:
I intentionally used lowercase “k”, the [SI prefix for “kilo”](https://en.wikipedia.org/wiki/Kilobyte).

The [Danish Orthographical Dictionary](https://en.wikipedia.org/wiki/Retskrivningsordbogen), the authoritative source for Danish spelling, explicitly mentions that [“kilobyte” is abbreviated “kB”](https://dsn.dk/soegning/?soegeord=kilobyte).

The Danish orthography is not always in sync with scientific standards (e.g. “deciliter” is abbreviated “dl” according to the Danish Language Council, but “dL” according to SI), but with respect to “kilobyte” they agree.

PR #1109, opened 2023-11-03
Add ordinalization for locales with simple pattern

c960657:
According to CLDR, a lot of locales use the same ordinalisation pattern for all numbers:
https://github.com/unicode-org/cldr/blob/12380e8/common/supplemental/ordinals.xml#L16-L18

Of these locales, the majority inherit from the simple rules defined in root.xml:
https://github.com/unicode-org/cldr/blob/12380e8/common/rbnf/root.xml#L698-L703

The rule just says that a number is ordinalised by adding a trailing period (aka [_ordinal dot_](https://en.wikipedia.org/wiki/Ordinal_indicator#Ordinal_dot)), just like we currently do for German:
https://github.com/svenfuchs/rails-i18n/blob/master/rails/ordinals/de.rb

In fact, the rules for German are so simple, so we don't have to use Ruby files but can just use YAML files like in `rails/locales/`. Also, there was a bug in the existing definitions for `de-AT`, `de-CH` and `de-DE` – they all define rules for `de`, not for each region-specific locale.

Based on the tool from #1107 I used `Cldr::Export.export(components: [Rbnf], …)` to generate `rbnf.yml` for all rails-i18n locales whose CLDR counterpart includes an rbnf.xml. In order to get proper inheritance from root, I applied this patch: https://github.com/ruby-i18n/ruby-cldr/pull/257.

I then found all rails-i18n locales whose ordinal rules were identical to the rules defined in root.xml. For each of these locales I generated an identical file in `rails/ordinals`.


spec/unit/ordinals_spec.rb
```diff
@@ -5,9 +5,10 @@
   # Mock Rails app in order to trigger the Railtie
   let(:app) { double :app, config: config }
   let(:config) { double :config, eager_load_namespaces: [], i18n: I18n, rails_i18n: RailsI18n }
+  let(:period_locales) { %w[be bs cs da de eo et fa fi hr hu is ka lb lt lv mk nb ne nn pl sk sl sq sr sw tr] }
 
   before do
-    I18n.available_locales = %w[fr en fr-CA fr-CH fr-FR gd]
+    I18n.available_locales = %w[fr en fr-CA fr-CH fr-FR gd]+ period_locales
```
sunny, line 11:
```suggestion
    I18n.available_locales = %w[fr en fr-CA fr-CH fr-FR gd] + period_locales
```

sunny (approved):
👏🏻

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## unreleased
 
+- Add ordinalization for many locales (be, bs, cs, da, eo, et, fa, fi, hr, hu, is, ka,
+  lb, lt, lv, mk, nb, ne, nn, pl, sk, sl, sq, sr, sw, tr) #1109
```
sunny, line 4: We can probably add `de-AT, de-CH, de-DE` to the list.

c960657:
@pouyaemami But en-CA doesn't have a “simple” pattern (as defined in this PR), does it?

pouyaemami:
Ah, I think I missed the point of this PR. You're right, this isn't what I was hoping for. My issue is that I get this error: `Translation missing: en-CA.number.nth.ordinalized` but `en-CA` and `en` have the same ordinalized values and I assumed that this PR takes a known ordinalized value from a base locale and applies it to the region specific one.

PR #1111, opened 2023-11-07
[it] Fix `second` prompt

c960657:
I don't speak any Italian, but AFAICT the `second` prompt uses a plural form unlike the other prompts and unlike its English counterpart.


sunny (approved):

pama:
@c960657 could you fix the conflict? Thanks!

PR #1112, opened 2023-11-14
zh-TW, zh-HK, zh-YUE: Write abbr_day_names using traditional character

johnnyshields:
Although `周` is possible in these scripts, the character `週` is preferred.

`zh-CN` is not changed.


sunny (approved):
Thanks!

Can you add an entry to the CHANGELOG as well? 🙏🏻

johnnyshields:
@sunny done.

PR #1130, opened 2024-10-21
Update for Rails 8

Schwad:
Ref: https://github.com/svenfuchs/rails-i18n/pull/962

- Allow runtime railties 8.0.0 dependency.
- Bump version number to 8.0.0.
- Update Gemfile.lock.

### DIFF
--- a/rails-i18n.gemspec
+++ b/rails-i18n.gemspec
@@ -16,7 +16,8 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = '>= 1.8.11'
 
   s.add_dependency('i18n', '>= 0.7', '< 2')
-  s.add_dependency('railties', '>= 6.0.0', '< 8')
+  s.add_dependency('railties', '>= 6.0.0')
+
   s.add_development_dependency "rspec-rails", "~> 3.7"
   s.add_development_dependency "i18n-spec", "~> 0.6.0"
   s.add_development_dependency 'i18n-tasks', '~> 0.9.37'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## unreleased
+- Update dependencies to allow Rails 8+.
 - Update following locales:
   - Lithuanian (lt): Add missing keys (`x_years`, `in`, `model_invalid`, `required`, `round_mode`, `eb`, `pb`)
   - Portuguese (pt): Fixed `number.currency.format.format` and `helpers.submit.update` #1122
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     rails-i18n (7.0.9)
       i18n (>= 0.7, < 2)
-      railties (>= 6.0.0, < 8)
+      railties (>= 6.0.0)
 
 GEM
   remote: https://rubygems.org/


rails-i18n.gemspec
```diff
@@ -15,8 +15,8 @@ Gem::Specification.new do |s|
   s.require_path = 'lib'
   s.required_rubygems_version = '>= 1.8.11'
 
-  s.add_dependency('i18n', '>= 0.7', '< 2')
-  s.add_dependency('railties', '>= 6.0.0', '< 8')
+  s.add_runtime_dependency('i18n', '>= 0.7', '< 2')
+  s.add_runtime_dependency('railties', '>= 6.0.0', '< 9')
```
casperisfine, line 19: Should there be an upper constraint at all?

Schwad: I don't think so; was trying to introduce the least amount of friction to get this approved. 

If maintainers are okay with it I'd love to remove the upper constraint!

casperisfine: Just remove it. Upper constraints are almost always a bad idea.

casperisfine: This upper dependency require to do the same dance every few years:
  - https://github.com/svenfuchs/rails-i18n/commit/c797b4e3dae1a735ec57fb5d7bda49aa4a5bc2be
  - https://github.com/svenfuchs/rails-i18n/commit/306461b19cb367d44eba2fd4b210bda51a63347e
  - https://github.com/svenfuchs/rails-i18n/commit/85a8b273de40b425b75a933500e330c11f2f6e35

That prevent people from trying Rails release candidates etc. It's just a bad pattern.

Schwad: Done

chaadow: Exactly. Unfortunately some maintainers still believe it's safer to restrain the upper limit. which leads me to maintain a lot of forks every year

sunny (commented):
Could you add an entry to the CHANGELOG, too? 🙏🏻

casperisfine:
@sunny done.

rails-i18n.gemspec
ghiculescu: Thank you for removing this!!!

pama:
I get the point about upper constraints and understand that it might create difficulties for maintainers looking to update their code to the latest Rails version.

However, I want to ensure that the current release version aligns with the correct Rails I18n keys. I'm hesitant to drop the version constraint because we could run into issues with mismatched keys if we start using versions ahead of our release. It could also set a precedent for having to accept PRs with keys from previous versions, which I've been refusing.

I know we don’t have changes often, but they do happen, like the introduction of `zb:` one year ago.

ghiculescu:
I think having a minimum dependency is helpful. Rails 7 requires rails-i18n 7, Rails 8 requires rails-i18n 8, etc. That would solve @pama's issue?

casperisfine:
> I'm hesitant to drop the version constraint because we could run into issues with mismatched keys if we start using versions ahead of our release.

Well, restricting only the major doesn't make sense then, because Rails from a semver standpoint 7.2 is as much of a major as 8.0, and it just as likely to add a key.

But you are the maintainer, so do what you prefer. If you want to keep the constraints, it's up to you. But either way, can this be solved soon please? Rails 8.0 is in RC1 and this is making it harder for people to test it.

Schwad:
@pama if I reintroduce a constraint would it be possible to merge this please.

4e4c52:
Hey!

Have you come to a decision for the upper constraint?

Rails 8 is now released and this is blocking the upgrade. 😢

pama:
I cam merge this without a release and you can point it to the master branch. Will that work for you?

pama:
@Schwad could you reintroduce the constraint and fix the conflicts? I'm sorry for taking so long.

PR #1140, opened 2025-03-22
Serbian Cyrillic: Fix date format, February typo, and RSD unit

s-mage:
Hey, thanks for the library.

I noticed that date formatting is off for Serbian, so this PR changes that.

How to format a date:
https://www.opismenise.com/nedoumice/kako-se-pise-datum-u-srpskom-jeziku/

I also changed unit to `РСД` because that's what people use in cyrillic now. For international  use there is RSD which is also its' ISO code. 

Screenshot 2025-03-22 at 16 09 42

This is a screenshot from the app where I pay for electricity to prove the point. 

The next step would be adding a sr-el encoding and renaming sr to sr-ec, because Serbian allows two alphabets.


pama:
Can you also add an entry to the CHANGELOG please? 🙏🏻

s-mage:
done that @pama

PR #1142, opened 2025-04-04
Add proper plural forms to Croatian and Serbian decimal units

EldinGuzin:
This PR adds proper plural forms (one, few, many, other) to decimal units in both 
Croatian (`hr.yml`) and Serbian (`sr.yml`) translations.

Before this change, these translations were using simplified strings without proper 
plural support, unlike other Slavic language files (`bs.yml` for example) in the repository.

These changes ensure grammatically correct output when different numeric values 
are displayed.


pama:
Can you also add an entry to the CHANGELOG please? 🙏🏻

dixpac:
I think that same should be done for Serbian here https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/iso-639-2/scr.yml

PR #1146, opened 2025-05-12
Update eu.yml

microstudi:
Percentage in basque goes with the symbol first


sunny (approved):
Can you add an entry to the CHANGELOG as well? 🙏🏻

microstudi:
right!

PR #1149, opened 2025-08-02
add relative time translations for de, en, es, pt

glaszig:
this adds translations used by the `relative_time_in_words` helper added in rails/rails#55405 for

- english
- german
- spanish
- portuguese

those are the only languages i'm comfortable with. i'll leave the rest to you, world.


sunny (approved):
🚀

glaszig:
👀👀

PR #1153, opened 2025-09-09
Add Saudi Arabia Locale with Support for New Saudi Riyal Symbol

obahareth:
The new Saudi Riyal symbol scheduled to be released with Unicode 17.0

Raw symbol: `⃁`
Unicode codepoint: `U+20C1`

References:
- https://blog.unicode.org/2025/03/support-for-new-saudi-riyal-currency.html
- https://unicode-explorer.com/c/20C1

NOTE:
It probably isn't a good idea to merge or release this without widespread adoption of Unicode 17, I just made the PR to have it ready.

### DIFF
--- a/rails/locale/ar-SA.yml
+++ b/rails/locale/ar-SA.yml
@@ -1,0 +1,278 @@
+---
+ar:
+  activerecord:
+    errors:
+      messages:
+        record_invalid: 'فشل التحقّق من: %{errors}'
+        restrict_dependent_destroy:
+          has_one: لا يمكن حذف السجل لوجود سجل يعتمد عليه %{record}
+          has_many: لا يمكن حذف السجل لوجود سجلات تعتمد عليه %{record}
+  date:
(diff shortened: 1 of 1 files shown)


obahareth:
It looks like the current `ar.yml` file is using KWD (Kuwaiti Dinar) as the currency, it might be more apt to have it be `ar-KW.yml`? That might break things for a lot of folks though. I made a new file to avoid breaking anything.

Also I wasn't sure if the convention is to put the currency ISO symbol (e.g. SAR), or the actual new symbol.

pama:
@obahareth, since maintainers cannot validate every locale, we rely heavily on community contributions, and changes are generally accepted unless someone raises concerns.

I agree with keeping the new locale instead of changing ar.yml.

As for the currency unit, for now I would prefer using SAR instead of introducing a newly added Unicode symbol.

Feel free to update the PR if that works for you.

PR #1165, opened 2026-03-20
Add negative_format for currency in Dutch locale

frenkel:
This is for example also the same format as used by Excel in the Dutch locale.

afbeelding


sunny:
Could you add a small entry to the CHANGELOG as well?

frenkel:
Sure, I've added one.

PR #1166, opened 2026-03-24
Fix incorrect `negative_format` in ja.yml

willnet:
The definition of `ja.number.currency.format.negative_format` was added in https://github.com/svenfuchs/rails-i18n/pull/1151 , but the symbol order is wrong. The current format results in values like "-円100", but "-100円" is correct. The `ja.number.currency.format.format` defined just above already uses the correct order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- Fix incorrect `ja.number.currency.format.negative_format` definition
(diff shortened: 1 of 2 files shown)


sunny:
Thank you! Can you add an entry to the CHANGELOG as well? 🙏🏻

willnet:
@sunny done 😃

PR #1170, opened 2026-05-15
Cache Arabic pluralization ranges

tagliala:
Move the 3..10 and 11..99 lookup arrays into frozen constants instead of rebuilding them on every Arabic pluralization check.

The rule runs during translation lookups, so allocating those arrays on each call adds avoidable work in a hot path. Local benchmarking shows the rule path rising from about 1.7k i/s to 8.7k i/s while eliminating its per-call allocations.

---

> [!WARNING]
> Hello, 
>
> this PR and the code in this PR are human generated because I was checking Arabic translation rules and I've noticed this behavior and did some research for my own interest.
>
> I've double checked `git blame` and, at the best of my knowledge, it is not intentional to avoid constants. I also understand that `11..99` is quite big
>
> Please triple check if this is a good change and it is worth to commit or not. I understand that this adds some extra retained memory and slower initialization time
>
> The benchmark and the following investigation are AI generated

### Overview

The Arabic pluralization rule currently rebuilds the lookup arrays for 3..10 and 11..99 on every invocation.

That work is unnecessary. Pluralization runs during translation lookups, so these allocations happen on a hot path and add avoidable CPU and memory overhead whenever Arabic translations are resolved with a count.

The main drawback is a small readability and maintainability tradeoff, plus a tiny amount of permanently retained memory for the cached constants

This PR fixes that by hoisting those arrays into frozen constants and reusing them across calls. The change is small, behavior-preserving, and directly removes repeated allocation from a performance-sensitive path.

This is also consistent with patterns already present elsewhere in rails-i18n. Shared pluralization helpers such as OneFewOther, EastSlavic, and Romanian already cache lookup ranges in frozen constants instead of rebuilding them repeatedly. There are also still a few inline allocation sites in the main pluralization file, notably in the Lithuanian and Polish rules, so Arabic is not an isolated case.

### Why this change is worth merging

- It removes avoidable per-call allocations from the Arabic pluralization rule.
- It improves throughput in the rule itself substantially.
- It reduces allocation pressure in end-to-end translation lookups.
- It does not change pluralization semantics.
- It follows an optimization style that is already used in other pluralization helpers in this repository.

### Benchmarks

Benchmarks were executed on an Apple M5 Pro with Ruby 4.0.3.

Benchmark source code: https://gist.github.com/tagliala/170ca1c8f1be747b33fe3371b2155bed

#### IPS

| Workload | master | this branch | delta |
|---|---:|---:|---:|
| Rule only | 1,716.3 i/s | 8,679 i/s | 5.06x faster |
| Translations | 7,111.2 i/s | 8,624 i/s | 21.3% faster |

#### Benchmark.memory

| Workload | master | this branch | delta |
|---|---:|---:|---:|
| Rule only | 482.4 KB, 900 objects | 0 B, 0 objects | eliminated allocations |
| Translations | 70,192 B, 434 objects | 41,248 B, 380 objects | 41.2% less memory, 12.4% fewer objects |

#### Memory profiler

| Workload | master | this branch | delta |
|---|---:|---:|---:|
| Rule only | 22,500 alloc, 11.50 MiB | 0 alloc, 0 B | eliminated allocations |
| Translations | 10,802 alloc, 1.67 MiB | 9,452 alloc, 1005.16 KiB | 12.5% fewer allocs, 41.2% less memory |

### Codebase context

- Existing cached lookup constants:
  https://github.com/svenfuchs/rails-i18n/blob/d88d78de6317041b3e154b8d260c4cb35e1f8d36/lib/rails_i18n/common_pluralizations/one_few_other.rb#L4-L5
  https://github.com/svenfuchs/rails-i18n/blob/d88d78de6317041b3e154b8d260c4cb35e1f8d36/lib/rails_i18n/common_pluralizations/east_slavic.rb#L9-L12
  https://github.com/svenfuchs/rails-i18n/blob/d88d78de6317041b3e154b8d260c4cb35e1f8d36/lib/rails_i18n/common_pluralizations/romanian.rb#L6
- Remaining inline allocation sites:
  pluralization.rb

### Expected impact

This is a focused hot-path optimization with very low risk. The biggest gain is in the pluralization rule itself, where repeated array construction was dominating cost. The full translation path also benefits through lower allocation pressure and better throughput.

### DIFF
--- a/lib/rails_i18n/pluralization.rb
+++ b/lib/rails_i18n/pluralization.rb
@@ -1,6 +1,9 @@
 module RailsI18n
   module Pluralization
     module Arabic
+      FROM_3_TO_10 = (3..10).to_a.freeze
+      FROM_11_TO_99 = (11..99).to_a.freeze
+
       def self.rule
         lambda do |n|
           return :other unless n.is_a?(Numeric)
@@ -13,9 +16,9 @@ module RailsI18n
             :one
           elsif n == 2
             :two
-          elsif (3..10).to_a.include?(mod100)
+          elsif FROM_3_TO_10.include?(mod100)
             :few
-          elsif (11..99).to_a.include?(mod100)
+          elsif FROM_11_TO_99.include?(mod100)
             :many
           else
             :other


lib/rails_i18n/pluralization.rb
```diff
@@ -13,9 +16,9 @@ def self.rule
             :one
           elsif n == 2
             :two
-          elsif (3..10).to_a.include?(mod100)
+          elsif FROM_3_TO_10.include?(mod100)
```
sunny, line 19: Wouldn’t a more optimized version of this would be to use `#between?`?

```suggestion
          elsif mod100.between?(3, 10)
```

sunny (commented):
Thanks for looking into this!

lib/rails_i18n/pluralization.rb
tagliala: Hello, this opens a pandora box that I didn't want to open, but I was really curious about it 😅

`n.between?(3, 10)` != `(3..10).to_a.include?(n)`

`between?` or `Range#cover?` are much more efficient but... they use a continuous interval

With this implementation, `3.1` will be returned as `few` instead of `other`. Which may be correct, but then it should be done also for `:zero`, `:one`, `:two` and `:many`, potentially leaving other as a non-reachable branch.

Actually `gd` pluralization does use a continuous range and floor values, so it is not clear to me if this rule is universal

I've tried to ask AI how to deal with translations, and I've got some insights:

> Yes — the main explanation is that `Range#member?` is probably not safe here when the rule is meant to match discrete integers only.
> 
> For your `gd.rb` example, this line:
> 
> ```ruby
> elsif (3..19).member?(floorn)
> ```
> 
> is fine because it checks `floorn`, not `n`, so `2.1` becomes `2` first and won’t incorrectly match `3..19`.
> 
> The real concern is when code does range membership directly on a non-normalized numeric value, because in Ruby:
> 
> - `Range#member?` for numeric ranges behaves like `cover?`
> - so it accepts any numeric value within the interval
> - which is different from “is this exact integer one of the elements of the range?”
> 
> So:
> 
> - `(3..10).member?(3.5) #=> true`
> - but `(3..10).to_a.include?(3.5) #=> false`
> 
> That means for pluralization logic, the correct choice depends on intent:
> 
> - If the CLDR-style rule means “integer values from 3 to 10”, then `member?` on raw `n` is wrong.
> - If the rule is applied to an integer-normalized value like `n.floor`, `n.to_i`, or explicitly only when `n == n.to_i`, then `member?` is fine and avoids allocation.
> 
> For Arabic specifically, the answer depends on which operands the rule is supposed to use. In CLDR plural rules, decimals are often treated differently from integers, and many languages route non-integers to `:other`. So for `2.1`, `:other` is usually the more plausible expected result than `:few`, unless the implementation intentionally collapses decimals to integer buckets.
> 
> So the practical rule is:
> 
> - `range.member?(floorn)` or `range.cover?(floorn)` → OK
> - `range.member?(n)` on arbitrary numeric input → suspicious for plural rules
> - `range.to_a.include?(n)` → semantically discrete, but allocates
> - best minimal-change replacement for discrete integer rules is usually:
>   - normalize first, then use `cover?`, or
>   - require integer-ness, then use `cover?`
> 
> Example pattern:
> 
> ```ruby
> return :other unless n.is_a?(Numeric)
> 
> i = n.to_i
> return :other unless n == i
> 
> (3..10).cover?(i)
> ```
> 
> That preserves discrete-integer semantics without allocating an array.
> 
> So short answer:
> 
> - yes, your concern is valid
> - `Range#member?` is only harmless if the value has already been integer-normalized
> - in `rails/pluralization/gd.rb`, it looks intentional and safe because it uses `floorn`
> - for other languages, especially Arabic-style rules, using `member?` directly on raw decimals could be a bug, not an optimization
> 
> If you want, I can inspect the pluralization files in `svenfuchs/rails-i18n` and point out which languages are semantically safe, suspicious, or likely buggy.

sunny: Thanks for looking further into this rabbit-hole. So it _seems_ like checking if `is_a?(Integer)` would make it equivalent **and** faster?

tagliala: I did avoid that as well, because of x.9repeating floats proven equals to their integer representation:

```
> 4.99999999999999999.is_a?(Integer)
=> false

irb(main):005> (3..10).to_a.include?(4.99999999999999999)
=> true

> 5.0.is_a?(Integer)
=> false

> (5..10).to_a.include? 5.0
=> true
```

I was thinking to compare mod100 to its floor representation of mod100, but I was not confident to go in that direction. I can take a look

tagliala: ### bench

```rb
def between?(n)
  n%1 == 0 && n.between?(11, 99)
end

def cover?(n)
  n%1 == 0 && (11..99).cover?(n)
end

def include?(n)
  (11..99).to_a.include?(n)
end

%i[ips memory].each do |benchmark|
  Benchmark.send(benchmark) do |x|
    [10, 11, 55, 99].each do |n|
      x.report("between?(#{n})") { between?(n) }
      x.report("cover?(#{n})") { cover?(n) }
      x.report("include?(#{n})") { include?(n) }
    end

    x.compare!
  end
end
```

### results

```
Ruby version: 4.0.3

Comparison (IPS):
        between?(10): 20845189.8 i/s
          cover?(10): 19919804.2 i/s - 1.05x  (± 0.00) slower
        between?(99): 17186806.5 i/s - 1.21x  (± 0.00) slower
        between?(11): 17160709.5 i/s - 1.21x  (± 0.00) slower
        between?(55): 17046166.5 i/s - 1.22x  (± 0.00) slower
          cover?(11): 15694114.9 i/s - 1.33x  (± 0.00) slower
          cover?(55): 15693215.4 i/s - 1.33x  (± 0.00) slower
          cover?(99): 15674912.5 i/s - 1.33x  (± 0.00) slower
        include?(11):   955021.3 i/s - 21.83x  (± 0.00) slower
        include?(55):   871972.6 i/s - 23.91x  (± 0.00) slower
        include?(10):   829906.5 i/s - 25.12x  (± 0.00) slower
        include?(99):   809042.8 i/s - 25.77x  (± 0.00) slower

Comparison (Memory):
        between?(10):          0 allocated
          cover?(10):          0 allocated - same
        between?(55):          0 allocated - same
          cover?(55):          0 allocated - same
        between?(99):          0 allocated - same
          cover?(99):          0 allocated - same
        between?(11):          0 allocated - same
          cover?(11):          0 allocated - same
        include?(10):       1064 allocated - Infx more
        include?(55):       1064 allocated - Infx more
        include?(99):       1064 allocated - Infx more
        include?(11):       1064 allocated - Infx more
```

---

> [!Warning]
> With the Numeric precondition, yes — they are functionally equivalent for all practical inputs (Integer, Float, Rational, BigDecimal). The between? form is strictly preferable: same result, no allocation, clearer intent. The only theoretical divergence is Complex, where include? returns false/true and between? raises.

```
irb(main):006> (5..10).to_a.include? Complex(7, 0)
=> true
irb(main):007> Complex(7, 0) % 1
(irb):7:in '<main>': undefined method '%' for an instance of Complex (NoMethodError)
```

edit: maybe this is not an issue, because a complex would already being failing at `mod100 = n % 100`

tagliala: ## Benchmark environment

Benchmarks were run on an Apple M5 Pro with Ruby 4.0.3, using the same benchmark script and the same default settings on each branch.

Compared branches:

1. `master`
2. `feature/move-arabic-arrays-to-constants` (THIS PR)
3. [`experimental/arabic-between`](https://github.com/svenfuchs/rails-i18n/compare/master...tagliala:rails-i18n:experimental/use-between-for-arabic)

## High-level takeaway

- `experimental/arabic-between` has the fastest rule-only benchmark.
- `feature/move-arabic-arrays-to-constants` has the fastest end-to-end translation benchmark.
- Both optimized branches eliminate rule-only allocations entirely.
- Both optimized branches also converge to the same reduced translation memory profile.

## IPS

| Branch | Rule only IPS | Translations IPS |
|---|---:|---:|
| `master` | 1.697k i/s | 6.950k i/s |
| `feature/move-arabic-arrays-to-constants` | 8.671k i/s | 8.952k i/s |
| `experimental/arabic-between` | 11.187k i/s | 8.649k i/s |

## Benchmark.memory

| Branch | Rule only | Translations |
|---|---|---|
| `master` | 482 KB, 900 objects | 70 KB, 434 objects |
| `feature/move-arabic-arrays-to-constants` | 0 B, 0 objects | 41 KB, 380 objects |
| `experimental/arabic-between` | 0 B, 0 objects | 41 KB, 380 objects |

## Memory profiler

| Branch | Rule only | Translations |
|---|---|---|
| `master` | 22,500 allocs, 11.5 MiB | 10,802 allocs, 1.67 MiB |
| `feature/move-arabic-arrays-to-constants` | 0 allocs, 0 B | 9,452 allocs, 1005 KiB |
| `experimental/arabic-between` | 0 allocs, 0 B | 9,452 allocs, 1005 KiB |

## Relative to `master`

| Branch | Rule only | Translations |
|---|---|---|
| `feature/move-arabic-arrays-to-constants` | 5.11x faster, zero allocations | 28.8% faster, lower memory and fewer allocations |
| `experimental/arabic-between` | 6.59x faster, zero allocations | 24.4% faster, lower memory and fewer allocations |

## Summary comparison

| Branch | Main strength | Main tradeoff |
|---|---|---|
| `master` | Baseline implementation | Slowest and most allocation-heavy in every benchmark |
| `feature/move-arabic-arrays-to-constants` | Best full translation throughput | Rule-only benchmark is slower than `experimental/arabic-between` |
| `experimental/arabic-between` | Best rule-only throughput | Slightly slower than `feature/move-arabic-arrays-to-constants` for full translations |

tagliala:
@sunny I'll open a new PR to address Lithuanian as well.

Lithuanian performance is also heavily impacted by the missing cache arrays. While investigating `between?` was interesting, I’ll start by proposing this backward-compatible implementation that is already used in other languages.

PR #909, opened 2020-10-01
Partially revert french thin spaces from 889

benoittgt:
Thin spaces and non-breaking spaces are the best way to right french, but not using then do not means it is incorrect. Rarely this is used in webpages and a lot of people do not know how to type and use them.

Partially revert #889 

Related: https://github.com/svenfuchs/rails-i18n/pull/889#issuecomment-701965676

PR #1174, opened 2026-06-21
Update hu.yml – currency should have delimiter

HuBandiT:
In Hungarian, both non-currency numbers and currency numbers should have some kind of space as delimiter for digit grouping. (Well, with the difference from current behaviour that numbers  as parts of continuous text –  e.g. a price mentioned as part of a sentence – should have no delimiter up to four (4) digits, delimiter only starting from five (5) digits; examples: 100 Ft, 1000 Ft, 10 000 Ft, 100 000 Ft. But numbers lined up under each other – e.g. accounting – should perform digit grouping as implemented currently: 100 Ft, 1 000 Ft, 10 000 Ft, 100 000 Ft, so that the digit grouping is visually consistent among all rows.)

For some reason however, the delimiter for currency numbers was set to empty string, while for regular numbers it was a space. Looking into "git blame" of the affected lines, I found no trace of discussion for this discrepancy. So I assume it is a mistake or oversight.

Therefore I hereby propose to make the currency delimiter a space.

To be typographically correct (including not allowing breaks) I propose U+2007 FIGURE SPACE as delimiter, and U+00A0 NO-BREAK SPACE to join the currency unit. For consistency I also changed the delimiter of non-currency numbers to U+2007 FIGURE SPACE to match currency numbers.

(If you ever need to re-parse suchly formatted numbers, please use the Unicode-aware `[[:space:]]` character class to properly match these space characters, instead of the ASCII-only `\s` which will not match them. As a bonus, because two separate characters are used now, cutting off the currency unit vs. deleting the delimiter spaces to join the digits together for parsing becomes unambiguous.)

### DIFF
--- a/rails/locale/hu.yml
+++ b/rails/locale/hu.yml
@@ -138,15 +138,15 @@ hu:
   number:
     currency:
       format:
-        delimiter: ''
-        format: "%n %u"
+        delimiter: ' ' # U+2007 FIGURE SPACE – https://en.wikipedia.org/wiki/Figure_space
+        format: "%n %u" # U+00A0 NO-BREAK SPACE
         precision: 0
         separator: ","
         significant: true
         strip_insignificant_zeros: true
         unit: Ft
     format:
-      delimiter: " "
+      delimiter: ' ' # U+2007 FIGURE SPACE – https://en.wikipedia.org/wiki/Figure_space
       precision: 2
       separator: ","
       significant: true
@@ -167,7 +167,7 @@ hu:
         significant: true
         strip_insignificant_zeros: true
       storage_units:
-        format: "%n %u"
+        format: "%n %u" # U+00A0 NO-BREAK SPACE
         units:
           byte:
             one: bájt
--- a/rails/locale/hu.yml
+++ b/rails/locale/hu.yml
@@ -138,15 +138,15 @@ hu:
   number:
     currency:
       format:
-        delimiter: ' ' # U+2007 FIGURE SPACE – https://en.wikipedia.org/wiki/Figure_space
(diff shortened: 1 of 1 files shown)


pama:
This topic has already been discussed in issue #881, “currency shouldn’t use a simple space, use nbsp instead”, and in PR #889. The Unicode spaces introduced in #889 were later removed in #909 and #996. Please refer to those previous discussions for more details.

I agree that Hungarian currency amounts should use a space as the thousands delimiter if that is the correct Hungarian convention. A PR changing the currency delimiter from an empty string to a regular space will be accepted.

However, introducing U+2007 FIGURE SPACE and U+00A0 NO-BREAK SPACE goes beyond locale conventions and can introduce breaking changes by affecting rendering, line wrapping, copy/paste behaviour, parsing, CSS, and layout.

These decisions should remain under application control, as that has historically been the approach in this project.

For that reason, a PR introducing U+2007 or U+00A0 as defaults will not be merged.

HuBandiT:
Thank you for the references.

I created https://github.com/svenfuchs/rails-i18n/pull/1175 now.

In those discussions it is visible that there are two conflicting requirements voiced: 1. do not change behaviour to keep backwards compatibility (typographical correctness less important); 2. change behaviour to be typographically correct (backwards compatibility less important).

Do you think it would be possible to enhance rails-i18n to make this configurable so that both parties can be satisfied at the same time?

Keep the current behaviour (plain spaces) as default, so those relying on current behaviour do not have to change anything and nothing breaks for them, but also allow those interested in typographically more correct behaviour to opt-in into that.

Thank you for your consideration.

